PR #204, opened 2021-01-06
Reset service worker files with new create-react-app defaults

CarsonHerness:
create-react-app has a new template for service workers. This PR essentially is a reset on the service worker that enables the PWA, and it should address at least some of the PWA bugs we're facing.


public/manifest.json
```diff
@@ -18,7 +18,7 @@
       "sizes": "512x512"
     }
   ],
-  "start_url": "",
+  "start_url": ".",
```
akampen, line 21: Just wondering, what does this accomplish?

src/service-worker.ts
```diff
@@ -0,0 +1,74 @@
+/// <reference lib="webworker" />
```
akampen, line 1: Is this supposed to be commented out?

akampen: Is this file auto-generated?

src/serviceWorkerRegistration.ts
```diff
@@ -14,34 +14,68 @@ type Config = {
   onUpdate?: (registration: ServiceWorkerRegistration) => void;
 };
 
-function registerValidSW(swUrl: string, config?: Config): void {
+export function register(config?: Config): void {
```
akampen, line 17: Add docstring

src/serviceWorkerRegistration.ts
```diff
@@ -14,34 +14,68 @@ type Config = {
...
+        });
+      } else {
+        // Is not localhost. Just register service worker
+        registerValidServiceWorker(serviceWorkerUrl, config);
+      }
+    });
+  }
+}
+
+function registerValidServiceWorker(serviceWorkerUrl: string, config?: Config) {
```
akampen, line 51: Add doc string

src/serviceWorkerRegistration.ts
```diff
@@ -14,34 +14,68 @@ type Config = {
...
 
               // Execute callback
               if (config && config.onUpdate) {
                 config.onUpdate(registration);
               }
             } else {
-              // Everything has been pre-cached for offline use
+              // At this point, everything has been precached.
+              // It's the perfect time to display a
+              // "Content is cached for offline use." message.
```
akampen, line 78: Delete

src/serviceWorkerRegistration.ts
```diff
@@ -58,15 +92,14 @@ function registerValidSW(swUrl: string, config?: Config): void {
     });
 }
 
-function checkValidServiceWorker(swUrl: string, config?: Config): void {
+function checkValidServiceWorker(serviceWorkerUrl: string, config?: Config) {
```
akampen, line 95: Doc string

src/serviceWorkerRegistration.ts
```diff
@@ -91,40 +124,6 @@ function checkValidServiceWorker(swUrl: string, config?: Config): void {
...
-        });
-      } else {
-        // Is not localhost. Just register service worker
-        registerValidSW(swUrl, config);
-      }
-    });
-  }
-}
-
 export function unregister(): void {
```
akampen, line 127: Doc string

akampen (approved):

public/manifest.json
CarsonHerness: It was the default used by other manifest files, but I think we do actually want the old value here, so I changed it back.

src/service-worker.ts
CarsonHerness: This is actually the syntax for Triple Slash Directives (see [documentation](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html)). This was autogenerated, so I don't have the strongest understanding about what this does, but my current understanding is that it's used by the TypeScript compiler.

PR #231, opened 2021-01-13
Add loading to auth context

CarsonHerness:
Previously if a signed in user went to the admin page, it flashed the sign in screen before jumping to the authenticated user page. Now it shows a loading message while it fetches the authenticated information from Firebase, and then it shows the appropriate content.

To reproduce behavior: while signed in, refresh the admin page.


src/pages/Admin.tsx
```diff
@@ -12,9 +13,14 @@ const UnauthenticatedAdmin = React.lazy(
  * Component for the Admin page
  */
 const Admin: () => JSX.Element = () => {
-  const {isAuthenticated} = useAuth();
+  const {isAuthenticated, isLoading} = useAuth();
 
-  return isAuthenticated ? <AuthenticatedAdmin /> : <UnauthenticatedAdmin />;
+  if (isLoading) {
+    // TODO: Replace Loading text with loading component
```
akampen, line 19: Is this a TODO for future or was this supposed to be for this PR? If this is for future, make a Github issue so we don't forget to come back to this

akampen (changes requested):

src/pages/Admin.tsx
CarsonHerness: I created an issue for this, seen [here](https://github.com/CEHAT-Clinic/dashboard/issues/199)

akampen (approved):

PR #256, opened 2021-01-27
Implement Admin functionality

CarsonHerness:
This PR only allows the admin page component to determine if an authenticated user is an admin user or a general user. Currently, admin users can only be added manually to the admin/user document in Firestore, but the next PR will allow admin users to manage other admin users.


src/contexts/AuthContext.tsx
```diff
@@ -34,8 +37,27 @@ const AuthProvider: React.FC<Props> = ({children}: Props) => {
...
         setIsAuthenticated(true);
+
+        // Check if user is an admin user
+        firestore
+          .collection('admin')
+          .doc('users')
+          .get()
+          .then(doc => {
+            if (doc.exists) {
+              const userData = doc.data();
```
akampen, line 48: add inline comment (ex: userData is the info for all admin users)

src/contexts/AuthContext.tsx
```diff
@@ -34,8 +37,27 @@ const AuthProvider: React.FC<Props> = ({children}: Props) => {
...
+              if (userData) {
+                const adminUserIds: string[] = userData.userId ?? [];
+                if (adminUserIds.includes(user.uid)) setIsAdmin(true);
+              }
+            }
+          })
+          .catch(error => {
```
akampen, line 55: comment: in what case would this error be thrown?

akampen (changes requested):

akampen (approved):

PR #259, opened 2021-01-29
Update CSV function to take time parameters

CarsonHerness:


functions/src/download-readings.ts
```diff
@@ -0,0 +1,173 @@
...
+    startDate = new Date();
+    endDate = new Date(startDate);
+    endDate.setMonth(startDate.getMonth() - 1); // eslint-disable-line no-magic-numbers
+  }
+
+  // Initialize csv with headers
+  const headings = SensorReading.getCsvHeader();
+
+  const sensorList = (await firestore.collection('/sensors').get()).docs;
+  const readingsArrays = new Array<Array<string>>(sensorList.length);
```
acgrobman, line 90: I can't remember, have we been using this notation or string[][] notation?

functions/src/download-readings.ts
```diff
@@ -0,0 +1,173 @@
...
+ * @remarks
+ * The start and end values represent the non-inclusive interval of time to fetch
+ * readings for. The start and end values should be numbers that are the number
+ * of milliseconds since EPOCH.
+ *
+ * @example
+ * ```
+ * // Gets readings for the month of December (in PST)
+ * generateReadingsCsv(
+ *   {data: Buffer.from('{"start": 1606809599000, "end": 1609488000000}')}
```
acgrobman, line 48: Is there a reason it has to use the raw epoch as opposed to something like a timestamp object or something that abstracts away the weird numbers?

functions/src/index.ts
```diff
@@ -276,98 +246,20 @@ exports.calculateAqi = functions.pubsub
...
-    const readings = readingsArrays.map(stringArray => stringArray.join(''));
-    const readingsCsv = headings + readings.join('');
-
-    // Generate filename
-    // Put timestamp into human-readable, computer friendly for
-    // Regex removes all non-word characters in the date string
-    const dateTime = new Date().toISOString().replace(/\W/g, '_');
-    const filename = `pm25_readings_${dateTime}.csv`;
+const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
+  timeoutSeconds: 540,
```
acgrobman, line 255: Why? Comment might help here.

functions/src/download-readings.ts
CarsonHerness: Since the date is in JSON, I thought that milliseconds since EPOCH would be the most consistent way to represent dates, especially since JavaScript recommends not using the constructors for dates that use strings.

This [StackOverflow post](https://stackoverflow.com/questions/10286204/what-is-the-right-json-date-format) sort of describes my thoughts, based on the final advice given: "The easiest and probably most portable format is the timestamp containing milliseconds since 1970"

functions/src/download-readings.ts
CarsonHerness: We have used both throughout. Do you have an opinion about which to use?

acgrobman: Not really, I just wasn't sure if we had previously only been using one, we would want consitency but we are already inconsistent so 🤷‍♂️

functions/src/download-readings.ts
acgrobman: Yeah, definitely wouldn't want to do the string thing. I was just wondering if it would be possible to use the `FirebaseFirestore.Timestamp` type because I know that we'd been going back and forth to Firestore with that, but it's also possible I'm not understanding the use case here versus elsewhere.

CarsonHerness: Parameters are passed to PubSub functions through JSON, so it's not possible to use a TypeScript class here

acgrobman: That's annoying 😢

acgrobman (approved):

PR #299, opened 2021-02-08
Display all users and admin users for authenticated admin users

CarsonHerness:
The next PR will allow admin users to manage the users by adminifying users or removing users.


src/components/Admin/UserPage/ManageUsers.tsx
```diff
@@ -31,6 +92,45 @@ const ManageUsers: () => JSX.Element = () => {
...
+                .map((user, id) => (
+                  <Tr key={id}>
+                    <Td>{user.name}</Td>
+                    <Td>{user.email}</Td>
+                  </Tr>
+                ))}
+            </Tbody>
+          </Table>
+          <Table>
+            <TableCaption placement="top">All Current Users</TableCaption>
```
akampen, line 115: This caption is centered differently than "Current Admin Users", but I am not sure why... (it's closer to the right side of the screen)

src/components/Admin/UserPage/ManageUsers.tsx
```diff
@@ -31,6 +92,45 @@ const ManageUsers: () => JSX.Element = () => {
           textAlign="center"
         >
           <Heading>Manage Users</Heading>
+          <Table>
```
akampen, line 95: A few lines above this (maxWidth="500px"): 

Consider changing this to 800px or something larger than 500px so that on a computer the names don't take up 2 lines (then each row is skinnier, easier to see more data in my opinion.

acgrobman (approved):

src/components/Admin/UserPage/ManageUsers.tsx
```diff
@@ -31,6 +92,45 @@ const ManageUsers: () => JSX.Element = () => {
...
+                <Tr key={id}>
+                  <Td>{user.name}</Td>
+                  <Td>{user.email}</Td>
+                  <Td>{user.admin ? 'Yes' : 'No'}</Td>
+                </Tr>
+              ))}
+            </Tbody>
+          </Table>
+          {error && <Text textColor="red.500">{error}</Text>}
           <Button as="a" href="/admin" margin={1}>
```
acgrobman, line 134: What's the difference between as vs on click redirect?

src/components/Admin/UserPage/ManageUsers.tsx
CarsonHerness: I changed this to a heading so that we can control the styling more directly

src/components/Admin/UserPage/ManageUsers.tsx
CarsonHerness: I agree, I changed it to 1000px but let me know if you think a different maxWidth would look better

src/components/Admin/UserPage/ManageUsers.tsx
CarsonHerness: I thought it was more readable this way and it means I didn't have to import the Router here, instead I just have an a element that's styled as a Button

akampen (approved):

PR #309, opened 2021-02-11
Change auth context to use updated admin status field

CarsonHerness:


src/contexts/AuthContext.tsx
```diff
@@ -40,22 +40,20 @@ const AuthProvider: React.FC<Props> = ({children}: Props) => {
...
-                const adminUserIds: string[] = userData.userId ?? [];
-                if (adminUserIds.includes(user.uid)) setIsAdmin(true);
+                if (userData.admin) setIsAdmin(true);
               }
             }
           })
           .catch(error => {
-            // Error thrown upon failure to fetch the admin/users doc from Firestore
-            throw new Error(`Unable to fetch admin/users doc: ${error}`);
+            // Error thrown upon failure to fetch the users doc from Firestore
```
acgrobman, line 55: Does this error display to users or just in logs?

src/contexts/AuthContext.tsx
```diff
@@ -40,22 +40,20 @@ const AuthProvider: React.FC<Props> = ({children}: Props) => {
...
+          .doc(user.uid)
           .get()
           .then(doc => {
             if (doc.exists) {
-              // Get the document data that contains all admin userIds
               const userData = doc.data();
               if (userData) {
-                const adminUserIds: string[] = userData.userId ?? [];
-                if (adminUserIds.includes(user.uid)) setIsAdmin(true);
+                if (userData.admin) setIsAdmin(true);
```
acgrobman, line 50: Does caching isAdmin locally allow someone malicious to change their status? It probably doesn't cause we can do serverside checks too on any database calls but just wondering?

CarsonHerness: My understanding is that a user cannot modify internal React variables but also yes, the server-side Firestore rules restrict reading and writing Firestore docs based on what Firestore says about admin status

src/contexts/AuthContext.tsx
CarsonHerness: Just in logs

acgrobman (approved):

akampen (approved):

PR #337, opened 2021-02-19
language-button

acgrobman:
Implement language toggle feature. 

Differs from original design by placing the change language button for mobile into the menu bar and making it text based on mobile. Implemented this way for speed and also because we will be chakrifying the nav system soon.

Also switches hamburger icon to an icon from a png and relies on nav tag as a selector.


src/components/NavBar/NavigationBar.tsx
```diff
@@ -35,24 +44,37 @@ function NavigationBar(): JSX.Element {
...
-        {/* CEHAT logo */}
         {isNavVisible && (
-          <nav className="Nav">
+          <nav>
             <a href="/">{t('home')}</a>
             <a href="/health">{t('healthInfo')}</a>
             <a href="/about">{t('about')}</a>
             <a href="/admin">{t('admin')}</a>
+            <button id="changeLanguage" onClick={toggleLanguage}>
+              {/* Display globe on desktop, but text in menu bard on mobile */}
```
CarsonHerness, line 64: typo: bar not bard

src/components/NavBar/NavigationBar.tsx
```diff
@@ -1,18 +1,27 @@
...
-import menuIcon from './menuIcon.png';
 import {useTranslation} from 'react-i18next';
+import {FaBars, FaGlobeAmericas} from 'react-icons/fa';
 
 function NavigationBar(): JSX.Element {
+  // State for whether to use globe or text, must be kept
+  // separate because mobile should always be text, even
+  // when nav bar is hidden
+  const [isMobile, setIsMobile] = useState(
+    window.matchMedia('(max-width: 700px)')?.matches
```
CarsonHerness, line 12: Let's not use optional chaining here since if window.matchMedia doesn't exist, then it will default to null instead of a boolean like we'd want.

src/components/NavBar/NavigationBar.tsx
```diff
@@ -35,24 +44,37 @@ function NavigationBar(): JSX.Element {
     setIsNavVisible(!isNavVisible);
   }
 
-  const {t} = useTranslation('menu');
+  function toggleLanguage(): void {
```
CarsonHerness, line 47: Add docstring

src/components/NavBar/NavigationBar.tsx
```diff
@@ -1,18 +1,27 @@
...
+  // when nav bar is hidden
+  const [isMobile, setIsMobile] = useState(
+    window.matchMedia('(max-width: 700px)')?.matches
+  );
+
   // State of nav bar (always visible in large screen)
-  const [isNavVisible, setIsNavVisible] = useState(true);
+  const [isNavVisible, setIsNavVisible] = useState(!isMobile);
 
   function handleScreenChange(this: MediaQueryList): void {
```
CarsonHerness, line 18: (I know this wasn't your change but can you also add a docstring here please thank you)

src/components/NavBar/NavigationBar.tsx
```diff
@@ -1,18 +1,27 @@
 import React, {useState, useEffect} from 'react';
 import './NavigationBar.css';
 import cehatLogo from './CEHATLogo.png';
-import menuIcon from './menuIcon.png';
 import {useTranslation} from 'react-i18next';
+import {FaBars, FaGlobeAmericas} from 'react-icons/fa';
 
 function NavigationBar(): JSX.Element {
```
CarsonHerness, line 7: (I know this wasn't your change but can you also add a docstring here please thank you)

src/components/NavBar/NavigationBar.css
```diff
@@ -16,7 +16,7 @@ All conditional formatting for small/big screens is in this file */
   }
 
 /* Arranges links in the navigation bar w*/
```
CarsonHerness, line 18: I know the comments aren't your code, but there's a typo here

src/components/NavBar/NavigationBar.css
```diff
@@ -34,7 +34,7 @@ All conditional formatting for small/big screens is in this file */
   }
 
 /* Navigable link text gets larger when you hover*/
```
CarsonHerness, line 36: super nit: add a space after hover

CarsonHerness (commented):
IMO, I think we should make it more obvious on Desktop that Globe is a button you can use to toggle languages, but that may be a design decision to ask the liaisons about.

- a minor thing: on mobile, close the menu after the user clicks the change language button
- (if not too difficult): on mobile, add the globe icon to the left of the Change Language/Cambia Idioma language
- Design question/suggestion: We could just display the "Change language text" in the language that a user is changing to, right? So in English display "Cambia idoma" and vice versa
- Also, whenever the language is changed or the page is changed, i18next prints this to the console (and it says the source is i18next.js:22). Is there a way to turn this off? It's not necessarily harmful for the average user tho, so not a big priority

src/components/NavBar/NavigationBar.tsx
```diff
@@ -1,18 +1,29 @@
...
+    window.matchMedia('(max-width: 700px)')?.matches ?? false
+  );
+
   // State of nav bar (always visible in large screen)
-  const [isNavVisible, setIsNavVisible] = useState(true);
+  const [isNavVisible, setIsNavVisible] = useState(!isMobile);
 
+  /** Adjust UI for switching between narrow/mobile and wide/desktop modes */
   function handleScreenChange(this: MediaQueryList): void {
     if (this.matches) {
```
CarsonHerness, line 21: Add a comment about what this.matches means here

src/components/NavBar/NavigationBar.tsx
```diff
@@ -35,24 +46,44 @@ function NavigationBar(): JSX.Element {
     setIsNavVisible(!isNavVisible);
   }
 
-  const {t} = useTranslation('menu');
+  /** Toggle the language of the website between English and Spanish. Close menu bar after toggling if on mobile. */
+  function toggleLanguage(): void {
+    i18n.changeLanguage(i18n.language === 'en' ? 'es' : 'en');
+    if (isMobile) {
```
CarsonHerness, line 52: nit: you can put this on one line: `if (isMobile) toggleNav();`

src/components/NavBar/NavigationBar.tsx
```diff
@@ -35,24 +46,44 @@ function NavigationBar(): JSX.Element {
...
+          <nav>
             <a href="/">{t('home')}</a>
             <a href="/health">{t('healthInfo')}</a>
             <a href="/about">{t('about')}</a>
             <a href="/admin">{t('admin')}</a>
+            <button id="changeLanguage" onClick={toggleLanguage}>
+              {/* Display globe on desktop, but text in menu bar on mobile */}
+              {isMobile ? (
+                <span>
+                  <FaGlobeAmericas id="mobileGlobe" title={t('globeIcon')} />{' '}
```
CarsonHerness, line 73: If possible, add some padding or something such that the text and icon have the same horizontal axis

CarsonHerness (approved):

src/components/NavBar/NavigationBar.tsx
acgrobman: Do we do this anywhere else in our code? I feel like at least in my personal style guide, it's always better to do multi-line if statements because then if one later makes it longer, missing curly braces don't cause incorrect behavior.

src/components/NavBar/NavigationBar.tsx
acgrobman: I'm not quite sure what it does since it wasn't my code originally but I have added a comment explaining what I think it does

src/components/NavBar/NavigationBar.tsx
acgrobman: Keeping per small conversation right after Carson left on Wednesday (sorry Carson)

PR #361, opened 2021-02-24
Write to Circular Buffers

akampen:
This PR writes to two circular buffers. It also continues to record and read data the way we've always done it, so it should not change our current functionality.

pm25Buffer:
- has 3600 spots to correspond to (30 readings/hour)*(12 hours needed for nowcast)
- pm25BufferIndex keeps track of where the next spot to write to is
- pm25BufferStatus (exists = 1, inProgress = 2, doesNotExist = 3)

aqiBuffer:
- has 144 spots to correspond to (6 readings/hour)*(24 hours needed for graph)
- aqiBufferIndex keeps track of where the next spot to write is
- aqiBufferStatus (exists = 1, inProgress = 2, doesNotExist = 3)


functions/src/index.ts
```diff
@@ -15,6 +15,39 @@ const thingspeakUrl = (channelId: string) =>
   `https://api.thingspeak.com/channels/${channelId}/feeds.json`;
 const readingsSubcollection = (docId: string) => `/sensors/${docId}/readings`;
 
+interface pm25BufferElement {
```
CarsonHerness, line 18: Add doc comment

CarsonHerness, line 18: Move this content to a separate file

functions/src/index.ts
```diff
@@ -15,6 +15,39 @@ const thingspeakUrl = (channelId: string) =>
...
+const defaultPm25BufferElement: pm25BufferElement = {
+  timestamp: null,
+  channelAPm25: NaN,
+  channelBPm25: NaN,
+  humidity: NaN,
+  latitude: NaN,
+  longitude: NaN,
+};
+
+interface AQIBufferElement {
```
CarsonHerness, line 36: Add doc comment

functions/src/index.ts
```diff
@@ -15,6 +15,39 @@ const thingspeakUrl = (channelId: string) =>
...
+  timestamp: FirebaseFirestore.FieldValue | null;
+  aqi: number;
+}
+const defaultAQIBufferElement: AQIBufferElement = {
+  timestamp: null,
+  aqi: NaN,
+};
+
+enum bufferStatus {
```
CarsonHerness, line 45: Add doc comment

functions/src/index.ts
```diff
@@ -30,6 +63,17 @@ async function getThingspeakKeysFromPurpleAir(
...
 }
 
+/**
+ * Increment the buffer index, circling back to 0 if the index is at the last point
+ * @param index - current index
+ * @param length - length of the buffer array
+ */
+const incrementIndex = (index: number, length: number) => {
+  let newIndex = index + 1; // eslint-disable-line no-magic-numbers
+  newIndex = newIndex < length ? newIndex : 0; // eslint-disable-line no-magic-numbers
```
CarsonHerness, line 73: Use a mod here instead, like `const newIndex = (index + 1) % length;`

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
           humidity: reading.humidity,
           latitude: reading.latitude,
           longitude: reading.longitude,
         };
+        // Add to historical readings
         await readingsRef.add(firestoreSafeReading);
       }
 
+      // Add readings to the pm25 buffer
+      const docRef = firestore.collection('sensors').doc(knownSensor.id);
```
CarsonHerness, line 138: Why do we do another read here? Could we do this inside of the existing if statement using the same information from the data we're about to write to Firestore?

functions/src/index.ts
```diff
@@ -15,6 +15,39 @@ const thingspeakUrl = (channelId: string) =>
...
+  timestamp: FirebaseFirestore.FieldValue | null;
+  aqi: number;
+}
+const defaultAQIBufferElement: AQIBufferElement = {
+  timestamp: null,
+  aqi: NaN,
+};
+
+enum bufferStatus {
+  Exists = 1 /* eslint-disable-line no-magic-numbers */,
```
acgrobman, line 46: Can `bufferStatus` be zero indexed and then we don't need a magic number disabling? The [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/enums.html ) seems like it will auto-start at 0 if we leave out an assignment, but I'm not sure if there is a Firebase reason to start at 1. Also, can this be a `const  enum`, not that I fully understand the difference?

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
           longitude: reading.longitude,
         };
+        // Add to historical readings
         await readingsRef.add(firestoreSafeReading);
       }
 
+      // Add readings to the pm25 buffer
+      const docRef = firestore.collection('sensors').doc(knownSensor.id);
+
+      await docRef.get().then(doc => {
```
CarsonHerness, line 140: You shouldn't need to combine `await` and `then` here, just the `then` should work fine

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
+        if (doc.exists) {
+          const status = doc.get('pm25BufferStatus');
```
CarsonHerness, line 142: Does this field exist in the historical sensor reading database?

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
+            pm25BufferIndex = incrementIndex(
+              pm25BufferIndex,
+              pm25Buffer.length
+            );
+            // Update document with new buffer and index
+            docRef.update({
+              pm25BufferIndex: pm25BufferIndex,
+              pm25Buffer: pm25Buffer,
+            });
+          } else if (
```
CarsonHerness, line 159: Should this just be an `else`?

functions/src/index.ts
CarsonHerness: After reading more code, is this supposed to be a different collection?

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
+        if (doc.exists) {
+          const status = doc.get('pm25BufferStatus');
+          if (status === bufferStatus.Exists) {
+            // If the buffer exists, update normally
+            let pm25BufferIndex = doc.get('pm25BufferIndex');
+            const pm25Buffer = doc.get('pm25Buffer');
+            // Add to circular buffer
+            pm25Buffer[pm25BufferIndex] = firestoreSafeReading;
+            // Increment index
+            pm25BufferIndex = incrementIndex(
```
CarsonHerness, line 150: instead of using a helper function, just use the one-liner I put above

functions/src/index.ts
```diff
@@ -242,13 +328,52 @@ exports.calculateAqi = functions.pubsub
...
+      await firestore.collection('current-reading').doc('pm25').set({
+        lastUpdated: FieldValue.serverTimestamp(),
+        data: currentData,
+      });
 
-    await firestore.collection('current-reading').doc('pm25').set({
-      lastUpdated: FieldValue.serverTimestamp(),
-      data: currentData,
-    });
+      /// / -------- Update the AQI Circular Buffer -------- ////
```
CarsonHerness, line 342: nit: use the `/* ---- */` style of comment here

functions/src/index.ts
```diff
@@ -30,6 +63,17 @@ async function getThingspeakKeysFromPurpleAir(
   return new PurpleAirResponse(purpleAirApiResponse);
 }
 
+/**
+ * Increment the buffer index, circling back to 0 if the index is at the last point
+ * @param index - current index
+ * @param length - length of the buffer array
+ */
+const incrementIndex = (index: number, length: number) => {
```
acgrobman, line 71: I'm not sure I can think of something better, but this function name makes me feel like it is going to modify the number, not just return the number to use

functions/src/index.ts
```diff
@@ -242,13 +328,52 @@ exports.calculateAqi = functions.pubsub
...
-    await firestore.collection('current-reading').doc('pm25').set({
-      lastUpdated: FieldValue.serverTimestamp(),
-      data: currentData,
-    });
+      /// / -------- Update the AQI Circular Buffer -------- ////
+
+      // Document reference for current sensor
+      const sensorDocRef = firestore.collection('/sensors').doc(docId);
+
+      await sensorDocRef.get().then(doc => {
```
CarsonHerness, line 347: Again, you shouldn't have to have both the await keyword and then, here's a [blog post](https://www.smashingmagazine.com/2020/11/comparison-async-await-versus-then-catch/) that I found helpful

functions/src/index.ts
```diff
@@ -242,13 +328,52 @@ exports.calculateAqi = functions.pubsub
...
+          const status = doc.get('aqiBufferStatus');
+          if (status === bufferStatus.Exists) {
+            // The buffer exists, proceed with normal update
+            let aqiBufferIndex = doc.get('aqiBufferIndex');
+            const aqiBuffer = doc.get('aqiBuffer');
+
+            // Update circular buffer
+            aqiBuffer[aqiBufferIndex] = aqiBufferData;
+            // Increment buffer index
+            aqiBufferIndex = incrementIndex(aqiBufferIndex, aqiBuffer.length);
```
CarsonHerness, line 358: use one-liner instead of helper function

functions/src/index.ts
```diff
@@ -242,13 +328,52 @@ exports.calculateAqi = functions.pubsub
...
+            // Update document
+            sensorDocRef.update({
+              aqiBufferIndex: aqiBufferIndex,
+              aqiBuffer: aqiBuffer,
+            });
+          } else if (
```
CarsonHerness, line 364: Like above, should this be an `else`?

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
   timeoutSeconds: 540,
 };
 
+/**
+ * This function populates the given sensor doc with a default circular buffer
+ * for eitherAQI or PM2.5
```
CarsonHerness, line 387: add space between "either" and "AQI" and add space between "PM" and "2.5"

functions/src/index.ts
acgrobman, line 140: This mixture of `async`/`await` notation and `.then()` chaining on the same line feels a bit confusing to me.

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
...
+ * @param aqiBuffer - true if AQI buffer, false if PM25 buffer
+ * @param docRef - document reference for the sensor to update
+ */
+async function populateBuffer(
```
CarsonHerness, line 391: Change function name to reflect that it creates a *default* buffer

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
+      await docRef.get().then(doc => {
+        if (doc.exists) {
+          const status = doc.get('pm25BufferStatus');
+          if (status === bufferStatus.Exists) {
+            // If the buffer exists, update normally
+            let pm25BufferIndex = doc.get('pm25BufferIndex');
+            const pm25Buffer = doc.get('pm25Buffer');
+            // Add to circular buffer
+            pm25Buffer[pm25BufferIndex] = firestoreSafeReading;
+            // Increment index
```
acgrobman, line 149: This comment feels like it repeats the name of the function call and might not be necessary, but that's up to debate

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
...
+ * This function populates the given sensor doc with a default circular buffer
+ * for eitherAQI or PM2.5
+ * @param aqiBuffer - true if AQI buffer, false if PM25 buffer
+ * @param docRef - document reference for the sensor to update
+ */
+async function populateBuffer(
+  aqiBuffer: boolean,
+  docRef: FirebaseFirestore.DocumentReference<FirebaseFirestore.DocumentData>
+) {
+  let bufferSize = 0;
```
CarsonHerness, line 395: Since it looks like `bufferSize` is only defined and used within each if statement, don't define it here

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
           channelBPm25: reading.channelBPm25,
           humidity: reading.humidity,
           latitude: reading.latitude,
           longitude: reading.longitude,
         };
+        // Add to historical readings
         await readingsRef.add(firestoreSafeReading);
       }
 
+      // Add readings to the pm25 buffer
```
acgrobman, line 137: In retrospect after reading through the rest of this section, do we want to externalize this into a helper function we can just call?

functions/src/index.ts
acgrobman, line 342: Do we want to externalize this to the helper function as above? Most of the comments on this are likely to be very similar to the section on the readings buffer.

functions/src/index.ts
acgrobman, line 347: Similar confusion on the combo `async`/`await` and `.then()`

functions/src/index.ts
```diff
@@ -242,13 +328,52 @@ exports.calculateAqi = functions.pubsub
...
+        if (doc.exists) {
+          const status = doc.get('aqiBufferStatus');
+          if (status === bufferStatus.Exists) {
+            // The buffer exists, proceed with normal update
+            let aqiBufferIndex = doc.get('aqiBufferIndex');
+            const aqiBuffer = doc.get('aqiBuffer');
+
+            // Update circular buffer
+            aqiBuffer[aqiBufferIndex] = aqiBufferData;
+            // Increment buffer index
```
acgrobman, line 357: Similar comment on comment necessity

functions/src/index.ts
```diff
@@ -242,13 +328,52 @@ exports.calculateAqi = functions.pubsub
...
+            aqiBufferIndex = incrementIndex(aqiBufferIndex, aqiBuffer.length);
+            // Update document
```
acgrobman, line 359: Comment neccessity?

functions/src/index.ts
```diff
@@ -65,28 +109,67 @@ exports.thingspeakToFirestore = functions
...
+            let pm25BufferIndex = doc.get('pm25BufferIndex');
+            const pm25Buffer = doc.get('pm25Buffer');
+            // Add to circular buffer
+            pm25Buffer[pm25BufferIndex] = firestoreSafeReading;
+            // Increment index
+            pm25BufferIndex = incrementIndex(
+              pm25BufferIndex,
+              pm25Buffer.length
+            );
+            // Update document with new buffer and index
```
acgrobman, line 154: Comment necessity?

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
...
+  aqiBuffer: boolean,
+  docRef: FirebaseFirestore.DocumentReference<FirebaseFirestore.DocumentData>
+) {
+  let bufferSize = 0;
+  const bufferIndex = 0;
+  if (aqiBuffer) {
+    // 144 = (6 calls/hour * 24 hours) is the amount of entries we need to
+    // create a graph with 24 hours of data
+    bufferSize = 144; /* eslint-disable-line no-magic-numbers */
+    const aqiBuffer: Array<AQIBufferElement> = [];
```
acgrobman, line 401: If we know the size, I feel like we can be a bit more efficient by declaring size up front and using array methods to fill.
```suggestion
    const aqiBuffer: Array<AQIBufferElement> = Array(bufferSize).fill(defaultAQIBufferElement);
 ```
    
    
```

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
...
+    for (let i = 0; i < bufferSize; ++i) {
```
acgrobman, line 402:
```suggestion
```

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
...
+      aqiBuffer.push(defaultAQIBufferElement);
```
acgrobman, line 403:
```suggestion
```

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
...
+    }
```
acgrobman, line 404:
```suggestion
```

functions/src/index.ts
acgrobman, line 36: I've tended to find that style guides recomend only capitalizing the first letter of acronyms
```suggestion
interface AqiBufferElement {
```

functions/src/index.ts
```diff
@@ -15,6 +15,39 @@ const thingspeakUrl = (channelId: string) =>
...
+  humidity: NaN,
+  latitude: NaN,
+  longitude: NaN,
+};
+
+interface AQIBufferElement {
+  timestamp: FirebaseFirestore.FieldValue | null;
+  aqi: number;
+}
+const defaultAQIBufferElement: AQIBufferElement = {
```
acgrobman, line 40:
```suggestion
const defaultAqiBufferElement: AqiBufferElement = {
```

functions/src/index.ts
```diff
@@ -229,7 +312,10 @@ exports.calculateAqi = functions.pubsub
       const NOWCAST_RECENT_DATA_THRESHOLD = 2;
       const containsEnoughInfo =
         validEntriesLastThreeHours >= NOWCAST_RECENT_DATA_THRESHOLD;
-      // If there is not enough info, the sensor's data is not reported
+
+      // If there's enough info, the sensor's data is updated
+      // If there isn't, we send the default AQI buffer element
+      let aqiBufferData = defaultAQIBufferElement; // New data to add
```
acgrobman, line 318:
```suggestion
      let aqiBufferData = defaultAqiBufferElement; // New data to add
```

functions/src/index.ts
acgrobman, line 387: nit: missing space
```suggestion
 * for either AQI or PM2.5
```

functions/src/index.ts
```diff
@@ -257,6 +382,57 @@ const generateReadingsCsvRuntimeOptions: functions.RuntimeOptions = {
...
+          aqiBuffer: aqiBuffer,
+          aqiBufferStatus: bufferStatus.Exists,
+        });
+      }
+    });
+  } else {
+    // 3600 = (30 calls/ hour * 12 hours) is the amount of data needed for
+    // the AQI nowcast calculation
+    bufferSize = 3600; /* eslint-disable-line no-magic-numbers */
+    const pm25Buffer: Array<pm25BufferElement> = [];
```
acgrobman, line 419: Similar changes to above?

functions/src/index.ts
akampen: Adam's original code writes to a new document in the readings sub collection. This code writes to the fields in the sensor doc. I'm not really sure what this comment is getting at.

functions/src/index.ts
akampen: I originally didn't have the await and sometimes the function was finishing before the doc was being updated (so the emulated database didn't reflect any of the new changes). This fixed that problem.

functions/src/index.ts
akampen: No, it's a new field.

functions/src/index.ts
akampen: No, because if the status is bufferStatus.InProgress, we want to skip over it without initializing the new buffer.

functions/src/index.ts
akampen: Okay, I'll explore this and try to recreate the bug that led me to this decision after addressing the other initial comments.

functions/src/index.ts
akampen: Like above, no because we do not want to re-initialize the new buffer if it's already started being initialized.

functions/src/index.ts
akampen: Yep, good suggestion.

functions/src/index.ts
akampen: N/A, we got rid of this function

functions/src/index.ts
akampen: Carson mentioned this too. I did this in response to a bug, I will try to recreate the problem without the await.

functions/src/index.ts
akampen: Let's talk about it quickly during the dev session on 3/3

functions/src/index.ts
akampen: Let's talk about it quickly during the dev session on 3/3

functions/src/index.ts
akampen: Same response, will recreate bug (or bug will be solved).

functions/src/index.ts
```diff
@@ -148,14 +170,140 @@ exports.calculateAqi = functions.pubsub
...
+    }
+  });
+}
+
+/**
+ * This function adds an element to a sensor's PM2.5 buffer (if applicable)
+ * @param data - the element to add to the buffer
+ * @param docRef - the document reference for a particular sensor
+ */
+function addToPm25Buffer(
```
akampen, line 226: This code is similar to the last function. I think it's easier to read without being littered with if statements, but there is a lot of duplicate so if anyone else prefers that these become one function with if statements instead, I can do that.

acgrobman (approved):
Looks good to me. I am gonna let Carson weigh in on the two functions that are similar but gonna mark approve pending his feedback as well

functions/src/index.ts
```diff
@@ -161,14 +184,140 @@ exports.calculateAqi = functions.pubsub
...
+ * @param docRef - the document reference for a particular sensor
+ */
+function addToAqiBuffer(
+  data: AqiBufferElement,
+  docRef: FirebaseFirestore.DocumentReference<FirebaseFirestore.DocumentData>
+) {
+  docRef.get().then(doc => {
+    if (doc.exists) {
+      const status = doc.get('aqiBufferStatus');
+      if (status === bufferStatus.Exists) {
```
CarsonHerness, line 212: Add comment about not updating if bufferStatus is InProgress

functions/src/index.ts
```diff
@@ -161,14 +184,140 @@ exports.calculateAqi = functions.pubsub
...
+ * @param data - the element to add to the buffer
+ * @param docRef - the document reference for a particular sensor
+ */
+function addToPm25Buffer(
+  data: Pm25BufferElement,
+  docRef: FirebaseFirestore.DocumentReference<FirebaseFirestore.DocumentData>
+) {
+  docRef.get().then(doc => {
+    if (doc.exists) {
+      const status = doc.get('pm25BufferStatus');
```
CarsonHerness, line 246: Add comment about InProgress (like above)

functions/src/aqi-calculation/buffer.ts
```diff
@@ -0,0 +1,114 @@
...
+ */
+const defaultAqiBufferElement: AqiBufferElement = {
+  timestamp: null,
+  aqi: NaN,
+};
+
+/**
+ * Enumeration for the status of a buffer. This is necessary because if a buffer
+ * is 'InProgress', it is currently being initialized, so we don't start to
+ * initialize it again.
```
CarsonHerness, line 53: Add a quick comment that this is necessary since initializing the entire buffer is a time costly operation

functions/src/aqi-calculation/index.ts
```diff
@@ -1,22 +1,29 @@
 import * as functions from 'firebase-functions';
```
CarsonHerness, line 1: index.ts should remain in the src directory in order for the cloud functions to deploy--just helper functions and files can be moved to subfolders

functions/src/aqi-calculation/index.ts
```diff
@@ -52,36 +59,67 @@ exports.thingspeakToFirestore = functions
...
+          pm25BufferIndex: pm25BufferIndex,
+          pm25Buffer: pm25Buffer,
+        });
+      } else if (status === bufferStatus.DoesNotExist || status === undefined) {
+        // If the buffer does not exist, populate it with default values so
+        // it can be updated in the future
+        await sensorDocRef.update({
+          pm25BufferStatus: bufferStatus.InProgress,
+        });
+        populateDefaultBuffer(false, docId);
```
CarsonHerness, line 113: Add comment that this function updates the bufferStatus to Exists once completed

functions/src/aqi-calculation/index.ts
```diff
@@ -52,36 +59,67 @@ exports.thingspeakToFirestore = functions
...
-      // our program from getting blocked.
-      // Allocates up to a minute of the two minute runtime for delaying
-      const oneMinuteInMilliseconds = 60000;
-      const delayBetweenSensors = oneMinuteInMilliseconds / sensorList.length;
-      await new Promise(resolve => setTimeout(resolve, delayBetweenSensors));
+      // Add readings to the PM 2.5 buffer
+      const sensorDocRef = firestore.collection('sensors').doc(knownSensor.id);
+      const status =
+        knownSensor.data().pm25BufferStatus ?? bufferStatus.DoesNotExist;
+      // If the buffer status is In Progress we don't update the buffer at all
```
CarsonHerness, line 93: "... at all since the document is already being written to" or something along those lines

functions/src/aqi-calculation/index.ts
```diff
@@ -52,36 +59,67 @@ exports.thingspeakToFirestore = functions
...
-      // Delays the loop so that we hopefully don't overload Thingspeak, avoiding
-      // our program from getting blocked.
-      // Allocates up to a minute of the two minute runtime for delaying
-      const oneMinuteInMilliseconds = 60000;
-      const delayBetweenSensors = oneMinuteInMilliseconds / sensorList.length;
-      await new Promise(resolve => setTimeout(resolve, delayBetweenSensors));
+      // Add readings to the PM 2.5 buffer
+      const sensorDocRef = firestore.collection('sensors').doc(knownSensor.id);
+      const status =
+        knownSensor.data().pm25BufferStatus ?? bufferStatus.DoesNotExist;
```
CarsonHerness, line 92: Since we call knownSensor.data() so much here, let's create a separate variable for it

functions/src/aqi-calculation/index.ts
```diff
@@ -161,8 +208,34 @@ exports.calculateAqi = functions.pubsub
       };
 
       currentData[purpleAirId] = currentSensorData;
+
+      // Update the AQI circular buffer for this element
+      const sensorDocRef = firestore.collection('/sensors').doc(docId);
+
+      const status =
+        knownSensor.data().aqiBufferStatus ?? bufferStatus.DoesNotExist;
+      // If the buffer status is In Progress we don't update the buffer at all
```
CarsonHerness, line 217: Same comment as above

CarsonHerness: It's not that we don't update the buffer at all, we just don't start updating the buffer again

functions/src/aqi-calculation/index.ts
```diff
@@ -161,8 +208,34 @@ exports.calculateAqi = functions.pubsub
...
+        const aqiBuffer: Array<AqiBufferElement> = knownSensor.data().aqiBuffer;
+        aqiBuffer[aqiBufferIndex] = aqiBufferData;
+        /* eslint-disable-next-line no-magic-numbers */
+        aqiBufferIndex = (aqiBufferIndex + 1) % aqiBuffer.length;
+        await sensorDocRef.update({
+          aqiBufferIndex: aqiBufferIndex,
+          aqiBuffer: aqiBuffer,
+        });
+      } else if (status === bufferStatus.DoesNotExist) {
+        // Populate the buffer with default values, skip this sensor
```
CarsonHerness, line 230: By "skip this sensor", do you mean we don't add the most recent AQI to the buffer?

functions/src/aqi-calculation/buffer.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @param aqiBuffer - true if AQI buffer, false if PM 2.5 buffer
+ * @param docId - document ID for the sensor to update
+ */
+function populateDefaultBuffer(aqiBuffer: boolean, docId: string): void {
+  const docRef = firestore.collection('/sensors').doc(docId);
+  const bufferIndex = 0;
+  if (aqiBuffer) {
+    // 144 = (6 calls/hour * 24 hours) is the amount of entries we need to
+    // create a graph with 24 hours of data
+    const bufferSize = 144; /* eslint-disable-line no-magic-numbers */
```
CarsonHerness, line 73: is the eslint necessary here since it's just setting a variable to a number? I think that's allowed without disabling the linter

functions/src/aqi-calculation/index.ts
CarsonHerness: Also add comment that this costs an extra write

functions/src/aqi-calculation/index.ts
```diff
@@ -161,8 +208,34 @@ exports.calculateAqi = functions.pubsub
...
+        await sensorDocRef.update({
+          aqiBufferStatus: bufferStatus.InProgress,
+        });
+        populateDefaultBuffer(true, docId);
```
CarsonHerness, line 234: Like above, add comment that this costs an extra write

CarsonHerness (approved):

PR #374, opened 2021-02-26
Update collection and sensor name for download-readings

CarsonHerness:


functions/src/download-readings.ts
```diff
@@ -140,8 +140,8 @@ function generateAverageReadingsCsv(): void {
 
   // Current-reading collection has single doc
   const currentReadingDocRef = firestore
-    .collection('/current-reading')
-    .doc('pm25');
+    .collection('current-reading')
```
acgrobman, line 143: Does this correspond to some other change? I can't remember what this was about.

acgrobman (approved):

functions/src/download-readings.ts
CarsonHerness: Just for consistency across the codebase

PR #379, opened 2021-03-01
Add lastValidAqiTime and readingDocId to the current-reading doc

CarsonHerness:


functions/src/index.ts
```diff
@@ -88,6 +88,9 @@ exports.calculateAqi = functions.pubsub
   .schedule('every 10 minutes')
   .onRun(async () => {
     const sensorList = (await firestore.collection('sensors').get()).docs;
+    const previousDataDoc = (
```
acgrobman, line 91: Being another read, does this impact Amber's calculations?

acgrobman (approved):
Looks good, although I kinda forget what this is supposed to do/help us do

functions/src/index.ts
CarsonHerness: The next PR removes the call to the sensors collection (I had to add the readingDocId first to be able to do this), so it will reduce the number of reads every 2 minutes from O(num sensors) to O(1), which I'm not sure we were actually accounting for currently?

CarsonHerness:
> Looks good, although I kinda forget what this is supposed to do/help us do

It's mostly just necessary to be able to call the current-readings doc to get the list of sensors to make API calls for

PR #389, opened 2021-03-06
Add more data to each sensor doc

CarsonHerness:
This PR adds more data to each sensor's doc in the sensors collection

- Adds an `isActive` field that is checked before doing any calculations
- Adds a `lastSensorReadingTime` that updates whenever a new PurpleAir reading is added, and this reduces a read since the function no longer queries the database to check for a read with the same timestamp unless the `lastSensorReadingTime` is unknown
- `lastValidAqiTime` is added to the sensor doc in the sensors collection

Note that because of adding the if statement checking `isActive` that wraps most of the functions, it looks like many mores lines of code were changed than in reality

Once this PR is merged, I'll run the "merge duplicate sensors" cloud function one more time for the double sensor doc, since this PR starts using `isActive` before making API calls, and the second doc for that sensor is marked as inactive.


functions/src/index.ts
```diff
@@ -35,85 +36,107 @@ exports.thingspeakToFirestore = functions
...
-      let firestoreSafeReading = defaultPm25BufferElement;
-      if (
-        (
-          await readingsRef
-            .where('timestamp', '==', Timestamp.fromDate(reading.timestamp))
+        // If the lastSensorReadingTime field isn't set, query the database to find
+        // the timestamp of the most recent reading. If there are no readings in
+        // Firestore, then lastSensorReading is never changed from null
+        let lastSensorReadingTime: FirebaseFirestore.Timestamp | null =
+          sensorDocData.lastSensorReadingTime ?? null;
```
acgrobman, line 77: A bit confused by the `?? null` Wouldn't having a null of the left side then coalesce into null on the right? Why do we need to specially do this? Or is this to deal with other nullish things that aren't actually null?

acgrobman (approved):
Looks mostly good to me

functions/src/index.ts
CarsonHerness: The nullish coalescing operator checks for `undefined` or `null` values, and the timestamp can be `undefined` if there is no entry in the document, or `null` if it is defined in the document, but the most recent sensor reading time is unknown (which I use `null` to denote). This line just makes all undefined values `null` so that it's easier to process later, since all of these values are `Timestamp | null` rather than `Timestamp | null | undefined`.

functions/src/index.ts
```diff
@@ -129,120 +151,105 @@ exports.calculateAqi = functions.pubsub
...
-      let isValid = false;
-      let lastValidAqiTime: FirebaseFirestore.Timestamp | null = null;
+        // If there is not enough info, the sensor's status is not valid
+        if (containsEnoughInfo) {
+          // Only calculate the NowCast PM 2.5 value and the AQI if there is enough data
+          const nowCastPm25Result = NowCastConcentration.fromCleanedAverages(
+            cleanedAverages
+          );
+          currentSensorData.aqi = aqiFromPm25(nowCastPm25Result.reading);
+          currentSensorData.latitude = nowCastPm25Result.latitude;
```
akampen, line 207: I see that we update the latitude, longitude, valid, etc in the currentReadings doc, but we don't do it in the sensor doc. Is that a problem?

akampen: Ex: if a sensor moves, will it ever be updated in the sensor docs or only in current readings?

akampen (approved):

PR #400, opened 2021-03-10
Add modal to add a new sensor

CarsonHerness:
This PR is built on top of PR #380, and it allows admin users to add new sensors


CarsonHerness:
With the updated PurpleAir API call, an updated version of this PR will:

- Make an API call to PurpleAir to check that the sensor ID is correct, since we can get the sensor name and location from the ID (and the admin can confirm this is for the sensor they intended)
- The sensor will then be added to the PurpleAir API's group and have a sensor doc created for it
- The name form field should also be removed since this comes from PurpleAir

public/locales/es/administration.json
```diff
@@ -134,9 +134,17 @@
         "inactive": "inactivo",
         "changeActiveSensorError": "No se puede cambiar el status de activo del sensor ",
         "add": "Añadir un sensor nuevo",
+        "addComplete": "Sensor agregado",
+        "showAddHelp": "Mostrar ayuda",
+        "hideAddHelp": "Ocultar ayuda",
+        "addHelpPurpleAirId": "Para encontrar el PurpleAir ID, ir al mapa del PurpleAir, y haz click que quieres añadir. En el elemento emergente para el sensor, coloca el cursor sobre el 'Get This Widget', que se encuentra en la parte inferior. Esto mostrará un texto como este: '<div id='PurpleAirWidget_39165_module_...'. El PurpleAir ID es el numero después del PurpleAirWidget, entonces en este caso, el PurpleAir ID es 39165.",
+        "sensorAlreadyExists": "Un sensor con este PurpleAir ID ya existe",
+        "unableToAddSensor": "No se pudo agregar el sensor. Comprueba que el ID de PurpleAir sea correcto.",
```
akampen, line 142: super nit: put these in the same order that they are in the English file

acgrobman (approved):

src/components/Admin/UserPage/AddSensorModal.tsx
```diff
@@ -0,0 +1,369 @@
...
+        // Create a sensor doc for the added sensor
+        await firestore.collection('sensors').add({
+          name: sensorName,
+          purpleAirId: purpleAirIdNumber,
+          latitude: latitude,
+          longitude: longitude,
+          isActive: true,
+          isValid: false,
+          lastValidAqiTime: null,
+          lastSensorReadingTime: null,
```
akampen, line 187: Do we need to start a readings subcollection or does that happen automatically when we populate with the first historical reading

akampen (approved):

src/components/Admin/UserPage/AddSensorModal.tsx
CarsonHerness: This happens automatically

PR #444, opened 2021-03-18
Update to use new PurpleAir API

CarsonHerness:
All references to Thingspeak have been removed, and the way we decide to use or discard readings has been updated to accommodate the new PurpleAir API. As such, several files were changed or deleted, and the way we download data has also been updated. For readability, I also externalized more of the code from `index.ts`


functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -1,32 +1,370 @@
...
+  let meanPercentDifference: number | undefined = undefined;
+  let pm25: number | undefined = undefined;
+  let humidity: number | undefined = undefined;
+  let timestamp: Date | undefined = undefined;
+
+  data.forEach((value, index) => {
+    // Check the corresponding field name to determine how to handle the value
+    switch (fieldNames[index]) {
+      case 'sensor_index':
+        if (typeof value === 'number') id = value;
```
acgrobman, line 120: Would it be better to use `value instanceof number` as opposed to have a bunch of strings floating around here? Or does that only work on things in TS not from API calls?

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -1,32 +1,370 @@
...
+  return lastSensorReadingTime;
+}
+
+/**
+ * Converts the PurpleAir ID to a number, if necessary.
+ * @param id - PurpleAir ID stored as a string or number
+ * @returns the PurpleAir ID as a number
+ */
+function getPurpleAirId(id: string | number): number {
+  if (typeof id === 'string') return +id;
```
acgrobman, line 231: see above about instanceof vs typeof

functions/src/aqi-calculation/util.ts
```diff
@@ -1,69 +1,56 @@
...
- * - `channelBPm25` - PM2.5 reading for channel B
+ * - `pm25` - PM 2.5 reading for a sensor. This value is the average of the
+ *   PM 2.5 reading for channelA and channelB
  * - `humidity` - humidity reading for a sensor
+ * - `meanPercentDifference` - mean percent difference between the pseudo averages
+ *   of the readings for channelA and channelB, as calculated from the confidence
+ *   value returned by the PurpleAir API
  * - `timestamp` - the timestamp of the current reading
  */
 interface HistoricalSensorReading {
```
acgrobman, line 48: The word historical feels overloaded now: does it refer to the old format or the per reading data (i know it means the per reading data and the comment says that, but I'm not sure if there's a better name)

functions/src/download-readings.ts
```diff
@@ -103,16 +111,20 @@ async function generateReadingsCsv(
...
+    // TODO: Create separate files for different types of readings
     for (
       let readingIndex = 0;
       readingIndex < readingsList.length;
       readingIndex++
     ) {
-      const reading = SensorReading.fromFirestore(
-        readingsList[readingIndex].data()
-      );
+      const data = readingsList[readingIndex].data();
```
acgrobman, line 120: Will these changes interupt @akampen 's work on the download? How does the system tell if it is a new or old reading/whether the call to `data.channelAPm25` is going to succeed or not?

functions/src/aqi-calculation/calculate-aqi.ts
```diff
@@ -134,10 +134,18 @@ function aqiFromPm25(pm25Concentration: number): number {
  * updates the `current-readings` AQI data and AQI-validity status. Also updates
  * the aqiBuffer in each sensor doc.
  */
+
 async function calculateAqi(): Promise<void> {
-  const sensorList = (await firestore.collection('sensors').get()).docs;
+  // Initialize the currentData map
```
acgrobman, line 139: In some other files in this PR you use an actual Map object, why did we want to use the null object here? Is there a better comment then other than using the word Map?

functions/src/aqi-calculation/purple-air-response.ts
CarsonHerness: The internet says that `instanceof` is for objects and custom types while `typeof` is for primitives.

[Medium Article Source](https://levelup.gitconnected.com/comparing-the-javascript-typeof-and-instanceof-operators-8844f23633a)
[Stack Overflow Source](https://stackoverflow.com/questions/899574/what-is-the-difference-between-typeof-and-instanceof-and-when-should-one-be-used)

functions/src/aqi-calculation/util.ts
CarsonHerness: How about something like `StoredSensorReading`? `FirestoreSensorReading`?

functions/src/download-readings.ts
CarsonHerness: iirc, @akampen is currently testing downloads from the frontend. I created a bug to update this function though so that it will create two separate files with each reading type.

functions/src/aqi-calculation/calculate-aqi.ts
CarsonHerness: This is something you did way back when--this is the map that's stored in the current-readings collection in Firestore, so it's how we created a map for Firestore. Do you know of a different way to create a map for Firestore data?

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -1,32 +1,370 @@
...
+  // Only return a PurpleAirReading if all fields are defined
+  if (
+    id !== undefined &&
+    name !== undefined &&
+    latitude !== undefined &&
+    longitude !== undefined &&
+    meanPercentDifference !== undefined &&
+    pm25 !== undefined &&
+    humidity !== undefined &&
+    timestamp !== undefined
```
akampen, line 165: Can't we just use `id && name && etc..` because undefined is a falsey value? Or do we want to make sure that nulls are still okay?

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -1,32 +1,370 @@
...
+
+  for (const sensorDoc of activeSensorDocsSnapshot.docs) {
+    const readingsCollectionRef = firestore.collection(
+      readingsSubcollection(sensorDoc.id)
+    );
+
+    // Initialize the buffer element to the default value
+    let pm25BufferElement: Pm25BufferElement = defaultPm25BufferElement;
+
+    // Get the reading for this sensor
```
akampen, line 256: Not getting "the reading"? This is getting all of a sensor's current data. Change to something like "Get the data for this sensor"

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -1,32 +1,370 @@
...
+    const sensorDocData = sensorDoc.data() ?? {};
+    const purpleAirId = getPurpleAirId(sensorDocData.purpleAirId);
+
+    // If the lastSensorReadingTime field isn't set, query the readings
+    // collection to find the timestamp of the most recent reading.
+    const lastSensorReadingTime: FirebaseFirestore.Timestamp | null =
+      sensorDocData.lastSensorReadingTime ??
+      (await getLastSensorReadingTime(readingsCollectionRef));
+
+    // If the reading was not in the group query, then it did not receive a new
```
akampen, line 266: "If a reading for this sensor"

functions/src/download-readings.ts
akampen: Currently my code does not use this function, any changes are fine

akampen (approved):

functions/src/aqi-calculation/util.ts
CarsonHerness: I'm leaving as is, but when we document the website, we need to add that we use the term "historical" to describe the readings subcollection historical database for each sensor.

functions/src/aqi-calculation/purple-air-response.ts
CarsonHerness: I check undefined specifically because 0s are okay, which is also a falsey value.

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -1,32 +1,370 @@
...
+/**
+ * Converts PurpleAir's confidence value into percent difference
+ * @param confidence - confidence value from PurpleAir, between 0 and 100
+ * @returns meanPercentDifference value, or NaN if value lost in calculation.
+ * Any input that results in NaN means that the meanPercentDifference is high
+ * enough that the reading should be discarded anyways.
+ *
+ * @remarks
+ * PurpleAir's confidence values is calculated as follow. PurpleAir does not
+ * document what `pc` means.
```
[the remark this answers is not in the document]
CarsonHerness: Oh that's definitely what pc stands for!

PR #461, opened 2021-03-22
Refactor functions

CarsonHerness:
This PR is a precursor to updating to the new PurpleAir API. I found that I had done a large refactoring in the code that would be confusing to do with the API update, so I created this PR.

- Adds config for running functions locally with the new API key
- Updates .gitignore to include local config files
- Removes latitude and longitude logic from the cleaning the readings since we just use the most recent latitude and longitude value. This also removed the latitude and longitude from the PM2.5 buffer since we now include the most recent latitude and longitude in the sensor doc (which is what we had been using for the location anyways).
- Moves each of the helper functions (including the main cloud function) to their respective file in the aqi-calculation process: cleaning the readings or calculating the AQI
- I also switched every "PM 2.5" to "PM2.5" to be consistent with our front end (a reversal from my previous preference of PM 2.5, sorry lol)
- I believe this also fixes our bug with the buffers being updated with the same reading each time--I make a separate buffer variable and historical reading variable, and I fixed the get last sensor time function.


functions/package.json
```diff
@@ -3,14 +3,15 @@
...
+    "shell": "npm run config && npm run build && firebase functions:shell",
     "start": "npm run shell",
     "deploy": "firebase deploy --only functions",
     "logs": "firebase functions:log",
     "lint": "gts lint",
     "clean": "gts clean",
     "compile": "tsc",
-    "fix": "gts fix"
+    "fix": "gts fix",
+    "config": "firebase functions:config:get > .runtimeconfig.json"
```
acgrobman, line 14: Is our Github Action going to need to be modified to support the new configuration?

functions/src/aqi-calculation/cleaned-reading.ts
```diff
@@ -1,11 +1,226 @@
...
+    humidity: humiditySum / readings.length,
+  };
+
+  return averageReading;
+}
+
+/**
+ * Gets the hourly averages for the past 12 hours for a single sensor. If less than
+ * 90% of the readings are available for a time period, it leaves the data for that hour
+ * as undefined per the EPA guidance to ignore hours without 90% of the data.
```
akampen, line 42: In a recent PR I changed this value to 75% but did not update the docs

functions/src/aqi-calculation/cleaned-reading.ts
```diff
@@ -1,11 +1,226 @@
...
+ * @param readings - Array of non-null Pm25BufferElements
+ */
+function averageReadings(readings: Array<Pm25BufferElement>): BasicReading {
+  let channelAPm25Sum = 0;
+  let channelBPm25Sum = 0;
+  let humiditySum = 0;
+
+  for (const reading of readings) {
+    channelAPm25Sum += reading.channelAPm25;
+    channelBPm25Sum += reading.channelBPm25;
```
acgrobman, line 26: There is a compiler warning on this line

functions/src/aqi-calculation/cleaned-reading.ts
```diff
@@ -1,11 +1,226 @@
...
+        /* eslint-disable no-magic-numbers */
+        cleanedAverages[i] =
+          0.534 * averagePmReading - 0.0844 * reading.humidity + 5.604;
+        /* eslint-enable no-magic-numbers */
+      } else {
+        // If reading exceeds thresholds above
+        cleanedAverages[i] = Number.NaN;
+      }
+    } else {
+      // If less than 27 data points were available for that hour, the reading
```
acgrobman, line 151: This comment is now misleading since we are using 75% not 95%, sorry for not catching in the earlier PR where we swtiched

functions/src/index.ts
```diff
@@ -34,12 +30,14 @@ exports.thingspeakToFirestore = functions
   .runWith(thingspeakToFirestoreRuntimeOpts)
   .pubsub.schedule('every 2 minutes')
   .onRun(async () => {
-    const sensorList = (await firestore.collection('/sensors').get()).docs;
+    const sensorDocQuerySnapshot = await firestore
```
acgrobman, line 33: Is there a better name for this? It feels a little unclear, especially with all the extra stuff at the end and it obscuring an `isActive` requirement

functions/src/index.ts
```diff
@@ -34,12 +30,14 @@ exports.thingspeakToFirestore = functions
...
+      .collection('sensors')
+      .where('isActive', '==', true)
+      .get();
 
-    for (const sensorDoc of sensorList) {
+    for (const sensorDoc of sensorDocQuerySnapshot.docs) {
       const sensorDocData = sensorDoc.data();
-      const isActive = sensorDocData.isActive ?? true;
-      if (isActive && sensorDocData.purpleAirId) {
+      if (sensorDocData.purpleAirId) {
```
acgrobman, line 40: Why would a purple air ID be nullish?

functions/package.json
CarsonHerness: This is only necessary for running the code locally. The PurpleAir API keys are already deployed, and the `firebase functions:config:get` command line function returns the deployed configuration. For some reason, the emulators cannot use the deployed configuration, so you have to create a runtime file to use them.

functions/src/index.ts
CarsonHerness: It should be, but this is just an extra check if there's a doc in the `sensors` collection that doesn't have a `purpleAirId` field

akampen (approved):
I'm looking into the `channelBPm25Sum` warning now

functions/src/aqi-calculation/cleaned-reading.ts
CarsonHerness: Amber found it god bless

acgrobman (approved):

PR #467, opened 2021-03-22
Add Spanish translation corrections to sensors page

CarsonHerness:
This PR adds the translation corrections that Hillary added after I submitted the sensors PR, #380


acgrobman (approved):
Might need updating again after we switch APIs/what data we have but otherwise lgtm

PR #469, opened 2021-03-22
Implement about page

acgrobman:
Also relocates the admin panel to a link on the about page.


public/locales/en/about.json
```diff
@@ -1,3 +1,39 @@
 {
-    "heading": "About Us Page"
+    "pageHeading": "About Us",
+    "cehat": {
+        "heading": "South Gate CEHAT",
+        "part1": "The South Gate Community Environmental Health Action Team (CEHAT) is a group of local residents, business owners, community activists, and government employees whose mission is “to make a South Gate that is beautiful, safe, educated, represented, healthy, sustainable, prosperous, clean, and united.” The CEHAT was the first group in California to utilize the ",
```
CarsonHerness, line 5: Was the South Gate CEHAT the first in California or CEHATs in general?

CarsonHerness (approved):

public/locales/en/about.json
acgrobman: [The CDC article](https://blogs.cdc.gov/yourhealthyourenvironment/2016/02/11/community-environmental-health-activism-in-south-gate-los-angeles-county-ca/) just says “This is the first time the PACE EH has been used in California to identify a city’s environmental issues and develop an action plan to tackle them.” Is the current phrasing indicating the wrong thing?

CarsonHerness: I just wasn't sure if there were other "CEHATs", maybe say "The South Gate CEHAT was the first group ..."

But I think what you have is also fine

CarsonHerness: I just wasn't sure if there were other "CEHATs", maybe say "The South Gate CEHAT was the first group ..."

But I think what you have is also fine

akampen (approved):

public/locales/en/about.json
acgrobman: I think I'm gonna leave it just so there's not two sentences in a row that both start the same way.

PR #481, opened 2021-03-24
Add AQI graph to home page

akampen:
Adds the AQI graph component to the home page.


src/components/AqiGauge/AqiDial.tsx
```diff
@@ -80,20 +80,22 @@ const AqiDial: ({currentAqi}: DialProps) => JSX.Element = ({
...
-      <Text fontSize={14} mb={2}>
-        {t('moreInfo')}
-        <Link fontSize={14} color="#32bfd1" href="/health">
-          {' '}
-          {t('menu:healthInfo')}
-        </Link>
-      </Text>
-      <AqiLabel currentAqi={currentAqi} />
-    </Box>
+    <Flex height="100%" width="100%" justifyContent="center" align="center">
```
akampen, line 83: Just added this flex to center the dial

public/locales/es/graph.json
```diff
@@ -0,0 +1,10 @@
+{
+    "Monday":"Lu",
```
CarsonHerness, line 2: Maybe use these abbreviations instead, from [Wikipedia](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Spain#:~:text=In%20Spanish%2C%20there%20exists%20a,D%20means%20domingo%20(Sunday).)

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+  CartesianGrid,
+  ResponsiveContainer,
+} from 'recharts';
+import firebase, {firestore} from '../../firebase';
+import {useTranslation} from 'react-i18next';
+
+/**
+ * Interface for a single element in the AQI buffer
+ * timestamp - when this AQI was calculated and added to the buffer
+ * aqi - the current aqi value
```
CarsonHerness, line 17: "the current AQI value"

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+ * This component displays a graph for the last 24 hours of AQI data for the
+ * currently selected sensor.
+ */
+const AqiGraph: ({sensorDocId}: GraphProps) => JSX.Element = ({
+  sensorDocId,
+}: GraphProps) => {
+  const defaultYLimit = 300;
+  const hoursPerDay = 24;
+  const minutesPerHour = 60;
+  /* --------------- state maintenance variables ---------------  */
```
CarsonHerness, line 73: Nit: capitalize State

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+/**
+ * AQI Graph Display Component
+ * This component displays a graph for the last 24 hours of AQI data for the
+ * currently selected sensor.
+ */
+const AqiGraph: ({sensorDocId}: GraphProps) => JSX.Element = ({
+  sensorDocId,
+}: GraphProps) => {
+  const defaultYLimit = 300;
```
CarsonHerness, line 70: Add comment for what this means

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+            const currentDate = new Date();
+            const currentDay = currentDate.getDay();
+            const currentHour =
+              currentDate.getHours() +
+              currentDate.getMinutes() / minutesPerHour;
+            let maxAqi = 0; // Keep track of the highest AQI value to scale the graph
+            for (let i = 0; i < aqiBuffer.length; i++) {
+              // Get the index starting at the oldest reading and wrapping
+              // once we reach the end of the buffer
+              const index = (aqiBufferIndex + i) % aqiBuffer.length;
```
CarsonHerness, line 123: We could initialize a `bufferIndex` value outside of the loop and then just `++` the `bufferIndex` variable instead of repeatedly adding `i` to the `aqiBufferIndex`.

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+  }, [sensorDocId]);
+
+  // The graph doesn't include the hazardous category unless thee are data points
+  // that fall in that category. Any time the maximum y-axis value changes, this effect
+  // updates the styling for the graph accordingly
+  useEffect(() => {
+    if (yAxisLimit === defaultYLimit) {
+      // Don't include hazardous color and extra grid line
+      /* eslint-disable-next-line no-magic-numbers */
+      setYAxisTicks([0, 50, 100, 150, 200, 300]);
```
CarsonHerness, line 171: I think you could use your variables from above here, the `const good = 50`, etc. Maybe put those at the top of the component?

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+          <ScatterChart>
+            <CartesianGrid horizontalFill={horizontalFill} fillOpacity={0.2} />
+            <XAxis
+              type="number"
+              dataKey="x"
+              height={70}
+              name="Hours Ago"
+              tick={{dy: 25, dx: -30}}
+              /* eslint-disable-next-line no-magic-numbers */
+              ticks={[0, 6, 12, 18, 24]}
```
CarsonHerness, line 264: I think this is worth putting in a variable above, like `const hourTicks = [..`

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+        padding={1}
+      >
+        <ResponsiveContainer height={250} width="90%">
+          <ScatterChart>
+            <CartesianGrid horizontalFill={horizontalFill} fillOpacity={0.2} />
+            <XAxis
+              type="number"
+              dataKey="x"
+              height={70}
+              name="Hours Ago"
```
CarsonHerness, line 261: Translate

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+              dataKey="y"
+              name="AQI"
+              unit=""
+              label={{value: 'AQI', position: 'Right', dx: -20, rotate: 0}}
+              ticks={yAxisTicks}
+              interval={0}
+              padding={{top: 1}}
+              domain={[0, yAxisLimit]}
+            />
+            <Scatter name="Good" data={data.good} fill="#08E400" />
```
CarsonHerness, line 282: Translate these

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+  XAxis,
+  YAxis,
+  Scatter,
+  CartesianGrid,
+  ResponsiveContainer,
+} from 'recharts';
+import firebase, {firestore} from '../../firebase';
+import {useTranslation} from 'react-i18next';
+
+/**
```
CarsonHerness, line 14: Externalize the interfaces to another file like Util.ts or AqiGraph.types.ts or something

public/locales/es/graph.json
acgrobman: Agree. While I hadn't seen those abbreviations Carson found, they seem less awkward then these

public/locales/en/graph.json
```diff
@@ -0,0 +1,10 @@
+{
```
acgrobman, line 1: I feel like keys should start with lowercase names? If that is possible

src/components/AqiGauge/AqiDial.tsx
acgrobman: If the goal is to center the dial, couldn't the containing Box just become the Flex instead?

CarsonHerness (approved):

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -0,0 +1,311 @@
...
+      t('Wednesday'),
+      t('Thursday'),
+      t('Friday'),
+      t('Saturday'),
+      t('Sunday'),
+    ];
+    const hoursPerPeriod = 12;
+    const sunday = 6;
+    // Get the local time from the user's browser
+    const date = new Date();
```
acgrobman, line 209: I forget, will the Firestore queries automatically handle time zone conversion for us?

src/components/AqiGraph/AqiGraph.tsx
acgrobman: Do these appear on the screen? If not, I don't think it's necessary, but if so, I think these are already translated in a different namespace and you can just use that namespace in addition

acgrobman: Although I would lower case if they don't

src/components/AqiGraph/AqiGraph.tsx
akampen: I just switched to using the `i` from the for loop instead. This version is a legacy from the line graph where the order of the points is important.

src/components/AqiGraph/AqiGraph.tsx
akampen: I use `getDay()`, `getHours()`, and `getMinutes()` which get the those values in the **time zone of the browser**. So all the axis labels are in the local timezone. I also use these functions when I calculate how many hours ago each data point is from, so throughout this code I'm using the user's local time (even though it doesn't really matter for the hours-ago calculation above).

public/locales/en/graph.json
```diff
@@ -0,0 +1,17 @@
...
+    "saturday":"Sat",
+    "sunday": "Sun",
+    "noSensor": "Click a sensor to see its data!",
+    "time": "Time",
+    "good": "Good",
+    "moderate": "Moderate",
+    "sensitive": "Unhealthy For Sensitive Groups",
+    "unhealthy": "Unhealthy",
+    "very": "Very Unhealthy",
+    "hazardous": "Hazardous"
```
akampen, line 16: I added these here even though they exist in another namespace because I didn't want the AQI ranges attached ex: "Good (0 - 50)"

acgrobman: Are they also in the aqi table namespace without ranges?

acgrobman (approved):

PR #533, opened 2021-04-06
Deletion Phase 0c: Follow correct module boundry typing

acgrobman:
It appears that Google does require the `data: any` parameter in addition to the `context: CallableContext` parameter. We were experiencing no authentication because it was getting passed to the wrong object. Because `any` is the correct type, the linter must be disabled on the line. In the actual code later we will make sure only that no-explicit-any is disabled, but given the amount of disabling on the page for this test, we will let it slide for now.

### DIFF
--- a/functions/.eslintrc.json
+++ b/functions/.eslintrc.json
@@ -99,7 +99,8 @@
           "Subcollection",
           "breakpoint",
           "Breakpoint",
-          "uid"
+          "uid",
+          "purpleair"
         ]
       }
     ]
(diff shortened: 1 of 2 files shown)


CarsonHerness (commented):
Looks good, I think there is a way to avoid the `any` issue--we can require it to be a certain type here, the Firebase function is compatible with `any` type

Also I think you accidentally removed `purpleair` from the eslint file, that one should still be there

CarsonHerness (approved):

PR #535, opened 2021-04-06
Fix AQI Buffer Bug

CarsonHerness:
This moves the coefficient recalculation out of an if statement so that the coefficient is decreased for every hour (even if an hour does not have enough data).

I also added some comments and improved some database write code from when Amber and I were debugging.

This also removes an extraneous if-check for `isActive` since we now only get the docs that are active

### DIFF
--- a/functions/src/aqi-calculation/calculate-aqi.ts
+++ b/functions/src/aqi-calculation/calculate-aqi.ts
@@ -215,41 +215,35 @@ async function calculateAqi(): Promise<void> {
     currentData[currentSensorData.purpleAirId] = currentSensorData;
 
     // Update the AQI circular buffer for this element
-    const sensorDocRef = firestore.collection('sensors').doc(sensorDoc.id);
     const status = sensorDocData.aqiBufferStatus ?? bufferStatus.DoesNotExist;
 
-    switch (status) {
-      case bufferStatus.Exists: {
-        // The buffer exists, proceed with normal update
-        const aqiBuffer: Array<AqiBufferElement> = sensorDocData.aqiBuffer;
-        aqiBuffer[sensorDocData.aqiBufferIndex] = aqiBufferData;
+    const sensorDocUpdate = Object.create(null);
+    sensorDocUpdate.lastValidAqiTime = currentSensorData.lastValidAqiTime;
+    sensorDocUpdate.isValid = currentSensorData.isValid;
+    
+    if (status === bufferStatus.Exists) {
+      // The buffer exists, proceed with normal update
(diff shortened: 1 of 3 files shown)


acgrobman (approved):

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -139,69 +139,45 @@ async function purpleAirToFirestore(): Promise<void> {
     }
 
     // Add readings to the PM 2.5 buffer
+    const sensorDocUpdate = Object.create(null);
+
+    // Only add the most recent reading data if a new reading exists
+    if (reading) {
```
akampen, line 145: If a new reading does not exist, we still want to add a default buffer element (NaNs/nones). I think we need to add an else where we don't update the last reading time and stuff but still add an element to the buffer and increment the index.

akampen (changes requested):

functions/src/aqi-calculation/calculate-aqi.ts
```diff
@@ -153,105 +153,99 @@ async function calculateAqi(): Promise<void> {
...
-          currentSensorData.isValid = true;
-          currentSensorData.lastValidAqiTime = Timestamp.fromDate(new Date());
+    // If there is not enough info, the sensor's status is not valid
+    const NOWCAST_RECENT_DATA_THRESHOLD = 2;
+    if (validEntriesLastThreeHours >= NOWCAST_RECENT_DATA_THRESHOLD) {
+      // If the calculated AQI is infinity, then the sensor value is not valid
+      // Only calculate the NowCast PM2.5 value and the AQI if there is enough data
+      const nowCastPm25 = cleanedReadingsToNowCastPm25(cleanedAverages);
+      const aqi = aqiFromPm25(nowCastPm25);
+      if (Number.isFinite(aqi)) {
```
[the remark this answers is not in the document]
CarsonHerness: I can't remember either, but I think it was how it handled `NaN`s but I think we've now determined that we wouldn't get a `NaN` back from this function anyways

functions/src/aqi-calculation/purple-air-response.ts
CarsonHerness: This still happens--the `sensorDocUpdate` variable only controls what is in the sensor doc, not the buffer, which is updated in later lines. I updated this code to make this clearer though

akampen: That was definitely my bad, I thought it was all inside the if statement

akampen (approved):

PR #537, opened 2021-04-06
Download data for a specific sensor

akampen:
Extends the download data modal to allow specific sensors' data to be downloaded.

I've tested it, would recommend that you test it with 1-2 days of data as well (do something in April if you want all the fields to show up because of the switch to the new data format).


CarsonHerness (commented):
Currently I've tested functionality, and here are my thoughts so far:

- When naming a document that only has data for one sensor, I think we should append the sensor ID or sensor name onto the filename
- All of the fields were editable after I downloaded the document, so my brain thought I could edit the fields and download another file immediately. This is not the case, so I think we should hide the sensor ID and date fields after we've started to fetch the data, so a user has to close the modal to be able to download data again
- For selecting which sensor to get data for, I like the checkbox for "Download data for all sensors". For when the user is choosing for a specific sensor, I think it'd be easier if there was just a checkbox for all possible sensors with the sensor's name and PurpleAir ID that the user could then select (i.e. if an admin wanted the data for cehat 1 and cehat 2, they could check 2 boxes for each of those sensors)

src/components/Admin/UserPage/DownloadData/DownloadCSVButton.tsx
```diff
@@ -163,24 +201,33 @@ const DownloadCSVButton: ({
...
+          {readyForDownload && (
+            <Box>
+              <Text>{t('downloadData.whenReady')}</Text>
+              <CSVLink data={body} headers={header} filename={filename}>
+                <Button colorScheme="teal">{t('downloadData.download')}</Button>
+              </CSVLink>
+              <Button
+                marginTop={2}
+                colorScheme="red"
+                onClick={() => resetAllFields()}
```
CarsonHerness, line 223:
```suggestion
                onClick={resetAllFields}
```

src/components/Admin/UserPage/DownloadData/DownloadCSVButton.tsx
```diff
@@ -163,24 +201,33 @@ const DownloadCSVButton: ({
...
-            <CSVLink data={body} headers={header} filename={filename}>
-              <Button>{t('downloadData.download')}</Button>
-            </CSVLink>
-          </Box>
-        )}
-      </Box>
+      {!error && (
+        <Box paddingTop={2}>
+          {!fetchingData && !readyForDownload && (
+            <Button onClick={() => fetchData()}>
```
CarsonHerness, line 207:
```suggestion
            <Button onClick={fetchData}>
```

src/components/Admin/UserPage/DownloadData/DownloadCSVModal.tsx
```diff
@@ -130,6 +156,28 @@ const DownloadCSVModal: () => JSX.Element = () => {
                 </HStack>
                 <FormHelperText>{t('downloadData.endHelper')}</FormHelperText>
                 {/* End end date input fields */}
+                {/* Start Purple Air ID input fields */}
+                <Checkbox
+                  marginY={2}
+                  isChecked={downloadAll}
+                  onChange={event => {
+                    setDownloadAll(!downloadAll);
+                  }}
```
CarsonHerness, line 165:
```suggestion
                  onChange={() => setDownloadAll(!downloadAll)}
```

src/components/Admin/UserPage/DownloadData/Util.tsx
```diff
@@ -47,6 +59,64 @@ interface InputProps {
   setValue: React.Dispatch<React.SetStateAction<number>>;
 }
 
+/**
```
CarsonHerness, line 62: Request: please move the SensorInput stuff into a Util.tsx file in the main folder for ManageSensors because I'll use it in the delete sensor PR!

src/components/Admin/UserPage/DownloadData/Util.tsx
```diff
@@ -37,6 +38,17 @@ interface CSVButtonProps {
...
   endDay: number;
   error: string;
+  downloadAll: boolean;
+  paId: string;
+  resetSelectedSensor: () => void;
+}
+
+/**
+ * Props for the CSV Modal.
+ * sensors - the list of PA sensors to include in the drop down menu
```
CarsonHerness, line 48: nit: PurpleAir (not PA)

src/components/Admin/UserPage/DownloadData/Util.tsx
```diff
@@ -47,6 +59,64 @@ interface InputProps {
   setValue: React.Dispatch<React.SetStateAction<number>>;
 }
 
+/**
+ * Props for sensor input fields used in the `DownloadCSVModal`
```
CarsonHerness, line 63:
```suggestion
 * Props for sensor input fields used in `SensorInput`
```

CarsonHerness (approved):
I think you'll have to merge in main with the refactor PR, but overall this looks great!

src/components/Admin/ManageSensors/ManageSensors.tsx
```diff
@@ -351,4 +351,5 @@ const ManageSensors: () => JSX.Element = () => {
   }
 };
 
+export type {Sensor};
```
CarsonHerness, line 354: Maybe also move the `Sensor` interface into the Util file in the ManageSensors folder

public/locales/es/administration.json
```diff
@@ -151,13 +151,15 @@
             "invalidStart": "La fecha de inicio no es válida",
             "invalidEnd": "La fecha final no es válida",
             "startBeforeEnd": "La fecha de inicio debe ser anterior a la fecha final",
-            "validDates": "Las fechas son válidas"
+            "validDates": "Las fechas son válidas",
+            "emptyField": "Al menos un campo está vacío",
+            "yearDigits": "Los años deben tener 4 dígitos"
```
acgrobman, line 156: Ought vs. need?

public/locales/es/administration.json
```diff
@@ -179,6 +181,10 @@
             "oct": "oct",
             "nov": "nov",
             "dec": "dic"
-        }
+        },
+        "downloadAll": "Descargar datos de todos los sensores",
+        "whichSensor": "¿Para qué sensor le gustaría descargar datos?",
```
acgrobman, line 186: Should it be qué or cuál since they are choosing from a specific list of options

acgrobman (approved):

PR #554, opened 2021-04-08
Use centralized colors on health page

### LINKED ISSUE #520: Use centralized colors on Health page
Realized I missed a page.

acgrobman:
Also changes the text color of sensitive AQIs to white in the EPA color scheme because that's what we were using before. Open to feedback on reverting to black, but it looked kinda weird when I accidentally did it.

Closes #520


CarsonHerness (approved):
Looks good!

Issue #122, opened 2020-11-20
Add/remove sensors from admin interface


CarsonHerness:

acgrobman:
Did we finish this?

CarsonHerness:
No I'm doing remove rn

PR #556, opened 2021-04-08
Allow admins to delete sensors

CarsonHerness:
This PR enables admins to delete sensors. I've tested both the deletion map update and the sensor doc deletion on some dummy sensors, and it works.

Other changes included:

 - Some helper components were moved to Util files.
 - I changed the SensorInput to take in the docId of the sensor in question (and changed Amber's Download data as needed)

Problems:

- Now we have 4 buttons in the HStack, so on mobile it doesn't format. I will add a bug for this
- My computer cannot test that download still works. Please run this locally and try downloading data!


src/components/Admin/ManageSensors/DeleteSensorModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Component to delete a sensor. Includes a button to make the modal pop up
+ * and a modal to delete a sensor.
+ */
+const DeleteSensorModal: ({sensors}: DeleteSensorModalProps) => JSX.Element = ({
+  sensors,
+}: DeleteSensorModalProps) => {
+  // --------------- State maintenance variables ------------------------
+  const {isOpen, onOpen, onClose} = useDisclosure();
```
akampen, line 42: nit: Since you separated out "modal state", move this line under that comment

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+      deletionDocRef
+        .get()
+        .then(deleteDoc => {
+          const newDeletionMap =
+            deleteDoc.data()?.deletionMap ?? Object.create(null);
+
+          newDeletionMap[sensorDocId] = new Date();
+
+          deletionDocRef
+            .update({deletionMap: newDeletionMap})
```
akampen, line 113: did you mention something about using .set instead of .update? Feel free to disregard if that doesn't make sense, just remembering something from someone from this week.

public/locales/es/administration.json
```diff
@@ -200,5 +200,15 @@
         "whichSensor": "¿Cuál sensor le gustaría descargar datos?",
         "chooseSensor": "Escoge sensor",
         "anotherSensor": "Descargar datos para otro sensor"
+    },
+    "deleteSensor": {
+        "delete": "Eliminar sensor",
```
acgrobman, line 205: Borrar is what Hillary had me use for delete on data, so maybe borrar here too? But maybe eliminar makes more sense in the context of a sensor. Not sure.

public/locales/es/administration.json
```diff
@@ -200,5 +200,15 @@
...
+        "confirmDownload": "He descargado los datos que necesito.",
+        "acknowledgeDelete": "Reconozco que esta acción eliminará los datos históricos de este sensor.",
+        "cannotBeUndone": "Reconozco que esta acción no se puede deshacer.",
+        "deleteNote": "Eliminar un sensor lo eliminará del mapa y eliminará todos sus datos. Deberás volver a agregar el sensor para mostrar este sensor en el mapa nuevamente y recopilar los datos de este sensor. Si solo deseas dejar de mostrar este sensor temporalmente o de recopilar datos para este sensor, desactiva el sensor en su lugar.",
+        "confirmPurpleAirId": "Escriba el ID de PurpleAir del sensor para confirmar la eliminación",
+        "whichSensorToDelete": "¿Qué sensor te gustaría eliminar?",
```
acgrobman, line 211: Would it be cual because choosing from a list of prespecified options?

public/locales/en/administration.json
```diff
@@ -200,5 +200,14 @@
         "whichSensor": "Which sensor would you like to download data for?",
         "chooseSensor": "Choose sensor",
         "anotherSensor": "Download data for another sensor"
+    },
+    "deleteSensor": {
+        "delete": "Delete Sensor",
+        "confirmDownload": "I have downloaded any data I need.",
+        "acknowledgeDelete": "I acknowledge that this action will delete any historical data for this sensor.",
+        "cannotBeUndone": "I acknowledge that this action cannot be undone",
+        "deleteNote": "Deleting a sensor will remove it from the map and delete all of its data. You will need to re-add the sensor to show this sensor on the map again and to collect this sensor's data. If you only want to temporarily stop showing this sensor or collecting data for this sensor, deactivate the sensor instead.",
```
akampen, line 209: where does this appear in the code? I can't find it

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  size="md"
+                  onChange={event => {
+                    setConfirmPurpleAirId(event.target.value);
+                    setError('');
+                  }}
+                  value={confirmPurpleAirId}
+                />
+                <FormErrorMessage>{error}</FormErrorMessage>
+              </FormControl>
+              <Button
```
acgrobman, line 183: Should we add a left icon of FaTrash? The one on the delete old modal has it

src/components/Admin/ManageSensors/DownloadData/DownloadCSVButton.tsx
```diff
@@ -70,21 +74,23 @@ const DownloadCSVButton: ({
   ) {
     setFetchingData(true);
     const sensorsRef = firestore.collection('sensors');
-    let sensorDocs;
+    let sensorDocs: (
+      | firebase.firestore.QueryDocumentSnapshot<firebase.firestore.DocumentData>
```
acgrobman, line 78: What does the extra preceding bar on line 78 do?

acgrobman (commented):
Successfully downloaded data both for all sensors and just one sensor (although when downloading for all, the prompt to download for another sensor felt awkward)

akampen (approved):

public/locales/es/administration.json
CarsonHerness: I changed to borrar for consistency with your code--I'll change it back if Hillary recommends doing so

src/components/Admin/ManageSensors/DownloadData/DownloadCSVButton.tsx
CarsonHerness: Both lines are too long on their own, it's what the formatter did automatically

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
CarsonHerness: Yes, here I think we should use `update` since this doc should already exist and I want to avoid overwriting any data. This does mean that someone cloning this site would need to create this document for this to function properly--perhaps during bug fix/documentation time, we can write a script that would create all necessary collections and docs.

public/locales/en/administration.json
CarsonHerness: Oops good catch! I forgot to add it back when I changed to using a modal. I added it back above the sensor selector.

public/locales/es/administration.json
CarsonHerness: Yes I think so, I changed it

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
CarsonHerness: I agree, I added the same code you had in DeleteOldDataModal

CarsonHerness:
Closes #122

PR #563, opened 2021-04-08
Deletion Phase 2: Actually delete marked data

acgrobman:
Tested locally on a small sample set.

### DIFF
--- a/functions/src/delete-marked-readings.ts
+++ b/functions/src/delete-marked-readings.ts
@@ -1,0 +1,83 @@
+import {firestore} from './admin';
+import {readingsSubcollection} from './aqi-calculation/util';
+
+/**
+ * Deletes readings before date marked for sensors in the deletion list
+ */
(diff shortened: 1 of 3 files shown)


akampen (approved):

functions/src/delete-marked-readings.ts
```diff
@@ -0,0 +1,86 @@
...
+    const query = readingsCollectionRef
+      .where('timestamp', '<', deleteBeforeDate)
+      .limit(batchSize);
+
+    deleteSensorSubcollectionBatch(query, batchSize, sensorDocId);
+  }
+}
+
+/**
+ * Wrapper function for `deleteQueryBatch` that uses a custom resolve function that removes the sensor from the deletion map
```
CarsonHerness, line 28: Add comment that says this is only called if the query has size 0, meaning the deletion task is complete

CarsonHerness (approved):

Issue #452, opened 2021-03-19
Clear buffers when activating/deactivating


CarsonHerness:
To avoid having bogus readings when a sensor is activated or deactivated, the buffers need to be cleared

CarsonHerness:
Also add and remove the sensors from the PurpleAir group

CarsonHerness:
Fixed by #565

PR #565, opened 2021-04-08
Fix activate/deactivate bug

CarsonHerness:
This resets the buffers when a sensor is activated or deactivated.

This also fixes a bug with sensor deletion--when a sensor is deleted, it is now removed from the PurpleAir group 490, which we use to get data.

Additionally, this only allows deactivated sensors to be deleted. This helps address a possible bug where the getPurpleAir data function is running while the sensor's doc is being deleted--now this cannot happen.

Note that when a sensor is activated or deactivated, we do not modify the PurpleAir group 490 as a matter of simplicity. PurpleAir will continue to give us data in our request for deactivated sensors, but the backend code will ignore the reading.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -124,6 +124,7 @@
           "csv",
           "num",
           "axios",
+          "Axios",
           "camelcase",
           "monday",
           "tuesday",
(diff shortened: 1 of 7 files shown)


acgrobman (approved):

CarsonHerness:
Fixes #452

akampen (approved):

PR #570, opened 2021-04-09
Fix Buffer Bug

CarsonHerness:
The bug: basically, once the `defaultAqiBufferElement` was used once, it couldn't be used again. When we looped through all of the sensors and set the `aqiBufferElement` to the default value at the start before checking the readings, the variable would maintain its most recent value (in the previous loop iteration) rather than reset to the `defaultAqiBufferElement` object. Perhaps this has something to do with how TypeScript stores variables used in loops and how objects are stored/copied.

This PR also adds TODOs for where to add down messages--a group conversation about the best way to do this should occur before the TODOs turn into code.

This PR also cleans up the code some (for example, consistently using `Number.NaN` instead of `NaN`).

There was also a bug with how I filtered our invalid readings I believe--filter doesn't modify the array but rather it returns a new array with the filtered values.


akampen (approved):

functions/src/aqi-calculation/buffer.ts
```diff
@@ -15,13 +15,20 @@ interface Pm25BufferElement {
  * This is a default element for the pm25 buffer. When the buffer is initialized,
  * every element is a default element. When we get an invalid reading (i.e. a
  * reading that matches that of the last reading), we put this element in the
- * buffer
+ * buffer.
+ *
+ * @readonly
+ *
+ * @remarks
```
acgrobman, line 22: Maybe instead we should say to use copies of this as in `const objShallowCopy = {...Obj1};` rather than creating copies manually?

functions/src/aqi-calculation/buffer.ts
```diff
@@ -38,10 +45,17 @@ interface AqiBufferElement {
  * This is the default element for the AQI buffer. The buffer is initialized
  * with default elements at every index. When we don't have enough valid PM2.5
  * data to calculate AQI, we put a default element in the buffer.
+ *
+ * @readonly
+ *
+ * @remarks
```
acgrobman, line 51: Ditto above with copies as opposed to manual?

functions/src/aqi-calculation/buffer.ts
```diff
@@ -107,8 +121,4 @@ function populateDefaultBuffer(aqiBuffer: boolean, docId: string): void {
 
 export type {Pm25BufferElement, AqiBufferElement};
 
-export {
-  defaultPm25BufferElement,
-  defaultAqiBufferElement,
-  populateDefaultBuffer,
-};
+export {populateDefaultBuffer};
```
acgrobman, line 124: I think that approach would let us export those again?

functions/src/aqi-calculation/calculate-aqi.ts
```diff
@@ -189,8 +188,11 @@ async function calculateAqi(): Promise<void> {
     }
 
     // If there's enough info, the sensor's data is updated
-    // If there isn't, we send the default AQI buffer element
-    const aqiBufferData: AqiBufferElement = defaultAqiBufferElement; // New data to add
+    // If there isn't, we send the AQI buffer element with default values
```
acgrobman, line 191: And then this would be where we implement the copy?

functions/src/aqi-calculation/cleaned-reading.ts
```diff
@@ -38,7 +39,7 @@ function averageReadings(readings: Array<Pm25BufferElement>): BasicReading {
  * @param status - the status of the pm25Buffer (exists, does not exist, in progress)
  * @param bufferIndex - the next index to write to in the buffer
  * @param buffer - the pm25Buffer with the last 12 hours of data
- * @returns a BasicReading array of length 12 with the average PM2.5 value for each of the last 12 hours. If an hour lacks enough readings, then the entry for that hour is `undefined`.
+ * @returns a BasicReading array of length 12 with the average PM2.5 value for each of the last 12 hours. If an hour lacks enough readings, then the entry for that hour is null
```
acgrobman, line 42: Why do we prefer `null` to `undefined` here?

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -108,7 +103,12 @@ async function purpleAirToFirestore(): Promise<void> {
       : null;
 
     // Initialize the buffer element to the default value
-    let pm25BufferElement: Pm25BufferElement = defaultPm25BufferElement;
+    let pm25BufferElement: Pm25BufferElement = {
```
acgrobman, line 106: This would be another copy location?

functions/src/aqi-calculation/cleaned-reading.ts
CarsonHerness: Just to be more clear/intentional about what each array element contains. `undefined` refers to an object that hasn't been initialized while `null` means that an object is missing, so I think `null` is more accurate here. The beginning of [this blog post](https://dmitripavlutin.com/7-tips-to-handle-undefined-in-javascript/) sort of discusses the difference. What do you think?

acgrobman: Hmm. I think null is the right return value, but then I wonder if it wouldn’t be more appropriate to rather than fill with null and replace with actual, to initialize with undefined and replace with either actual or null. The outcome is the same, it’s just it feels like initializing with null goes against the idea that null is intentional absence.

functions/src/aqi-calculation/buffer.ts
CarsonHerness: I added a helper function that does this, lmk what you think

functions/src/aqi-calculation/cleaned-reading.ts
CarsonHerness: Sure that makes sense to me, I'll make the change

PR #577, opened 2021-04-10
Fix purpleAirToFirestore Update() error

### LINKED ISSUE #576: Error in purpleAirToFirestore cloud function
Error: Update() requires either a single JavaScript object or an alternating list of field/value 
pairs that can be followed by an optional precondition. At least one field must be updated.

It happens locally and is being logged in Firestore

CarsonHerness:
This PR adds a `lastUpdated` field for every update to a sensor's doc. This is necessary because the purpleAirToFirestore function was throwing errors if `update` was called but no fields were actually updated. Now, every time update is called, `lastUpdated` is set to the server timestamp.

This PR also:

- Fixes a bug in the type of the aqiBufferElement
- Fixes some documentation for the aqiBufferElement
- Adds TODOs for where a reason that a sensor is down can be determined
- Removes previous helper function to ensure that a `purpleAirId` is a number since all sensor docs have now been converted to using numbers, and all newly added sensors will also use a number

Closes #576

### DIFF
--- a/functions/src/aqi-calculation/purple-air-response.ts
+++ b/functions/src/aqi-calculation/purple-air-response.ts
@@ -12,7 +12,6 @@ import {
   readingsSubcollection,
   getReading,
   getLastSensorReadingTime,
-  getPurpleAirId,
 } from './util';
 
 /**
@@ -84,6 +83,8 @@ async function purpleAirToFirestore(): Promise<void> {
     .where('isActive', '==', true)
     .get();
 
+  console.log(activeSensorDocsSnapshot.docs);
+
   for (const sensorDoc of activeSensorDocsSnapshot.docs) {
     const readingsCollectionRef = firestore.collection(
       readingsSubcollection(sensorDoc.id)
@@ -91,7 +92,7 @@ async function purpleAirToFirestore(): Promise<void> {
 
     // Get the existing data from Firestore for this sensor
     const sensorDocData = sensorDoc.data() ?? {};
-    const purpleAirId = getPurpleAirId(sensorDocData.purpleAirId);
+    const purpleAirId: number = sensorDocData.purpleAirId;
 
     // If the lastSensorReadingTime field isn't set, query the readings
     // collection to find the timestamp of the most recent reading.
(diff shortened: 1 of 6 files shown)


functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -91,7 +90,7 @@ async function purpleAirToFirestore(): Promise<void> {
 
     // Get the existing data from Firestore for this sensor
     const sensorDocData = sensorDoc.data() ?? {};
-    const purpleAirId = getPurpleAirId(sensorDocData.purpleAirId);
+    const purpleAirId: number = sensorDocData.purpleAirId;
```
akampen, line 93: In this PR I don't see a place where we check for numbers (so I assume it happened in a recent PR), what new code ensured that we don't need to do this check anymore?

functions/src/aqi-calculation/purple-air-response.ts
```diff
@@ -133,7 +141,6 @@ async function purpleAirToFirestore(): Promise<void> {
         sensorDocUpdate.latitude = reading.latitude;
         sensorDocUpdate.longitude = reading.longitude;
         sensorDocUpdate.name = reading.name;
-        sensorDocUpdate.purpleAirId = reading.id;
```
akampen, line 136: is this removed just because it's redundant? Could PA ever change a PA ID?

functions/src/aqi-calculation/purple-air-response.ts
CarsonHerness: I manually checked all sensor docs in Firestore used numbers, so there's no longer a mix.

functions/src/aqi-calculation/purple-air-response.ts
CarsonHerness: This was my way of automatically changing the sensor docs to use numbers instead of strings for purpleAirId, but it's no longer necessary. The PurpleAir ID won't change in a sensor call

akampen (approved):

PR #579, opened 2021-04-12
Invalid sensors: changes for Dial Box & About page

akampen:
This PR does a couple things:
- Fixes translations for the home and about pages for the "invalid sensor/sensor down" stuff
- Small corrections on the English versions as well
- Uses "a las" when it's not 1 o'clock, uses "a la" when it is
- If a sensor has never given a valid reading, use different text in the dial box to reflect that.

### DIFF
--- a/public/locales/en/dial.json
+++ b/public/locales/en/dial.json
@@ -9,8 +9,10 @@
     "hazardous": "Hazardous (301+)",
     "invalid": {
         "lastTime": "This sensor is not up to date. The last time it reported an accurate AQI was ",
-        "at": " at ",
+        "atPlural": " at ",
+        "atSingular": " at ",
         "learnMore": ". To learn more about why a sensor might not be reporting, check out our ",
-        "aboutPage": "About Page."
+        "aboutPage": "About Page.",
+        "neverReported": "This sensor has never reported an accurate AQI"
     }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -63,6 +63,7 @@
         "skipWordIfMatch": [
           "sunday",
           "85vh",
+          "las",
           "dx",
           "dy",
           "aqi",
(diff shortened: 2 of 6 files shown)


public/locales/en/dial.json
```diff
@@ -9,8 +9,10 @@
     "hazardous": "Hazardous (301+)",
     "invalid": {
         "lastTime": "This sensor is not up to date. The last time it reported an accurate AQI was ",
-        "at": " at ",
+        "atPlural": " at ",
```
acgrobman, line 12: It may not be worth it, but I know i18next has built in plurals support. Is this using that?

public/locales/en/dial.json
```diff
@@ -9,8 +9,10 @@
...
+        "atSingular": " at ",
         "learnMore": ". To learn more about why a sensor might not be reporting, check out our ",
-        "aboutPage": "About Page."
+        "aboutPage": "About Page.",
+        "neverReported": "This sensor has never reported an accurate AQI"
```
acgrobman, line 16: Do we need a period here?

acgrobman (commented):
Looks pretty good to me, haven't had ability to test myself because on mobile. Not checking off just cause waiting for Carson to come through and review as well, but consider that I've approved pending his approval

CarsonHerness (approved):

public/locales/en/dial.json
akampen: We don't because the beginning of the next string starts w/ a period (to account for the fact that there's no period in the time-string).

public/locales/en/dial.json
akampen: It doesn't use that. I checked it out and it would end up being a very similar implementation to what we have rn, but if it's preferred that I do that, I can switch it.

PR #580, opened 2021-04-12
Bug fix: labels on AQI graph

akampen:
Not sure what happened here, seems like `.getDay()` returns 0 for Sunday, 6 for Saturday. Our code had implemented 0 for Monday, 6 for Sunday, which means all the labels started showing up as "undefined" (you can see it on the live site today).

This PR changes the code to reflect Sunday being 0, not 6.

Commentary: I feel like we would have noticed if this was always happening because then the graph would have always been off by a day. Not sure if/why it changed.


acgrobman (approved):
Strange but hopefully this fixes

PR #581, opened 2021-04-12
Added title to AQI graph

akampen:


public/locales/es/graph.json
```diff
@@ -7,5 +7,6 @@
     "saturday":"S",
     "sunday": "D",
     "noSensor": "¡Haz clic en un sensor para ver sus datos!",
-    "time": "Tiempo"
+    "time": "Tiempo",
+    "graphTitle": "(Air Quality Index, AQI) en las últimas 24 horas"
```
acgrobman, line 11: I would recommend using the translation for AQI on this as well as the dial?  That is the Índice de...

acgrobman (changes requested):

public/locales/es/graph.json
CarsonHerness: I think we could just use AQI here too since AQI is defined in the gauge component

src/components/AqiGraph/AqiGraph.tsx
```diff
@@ -198,7 +198,9 @@ const AqiGraph: ({sensorDocId}: GraphProps) => JSX.Element = ({
         justifyContent="center"
         align="center"
         padding={1}
+        flexDir="column"
       >
+        <Heading fontSize="md">{t('graphTitle')}</Heading>
```
CarsonHerness, line 203: Just my opinion
```suggestion
        <Heading fontSize="lg" marginBottom={2}>
          {t('graphTitle')}
        </Heading>
```

CarsonHerness (approved):

public/locales/es/graph.json
acgrobman: I’m cool with that too, but just am not a fan of the parenthetical only

acgrobman (approved):

PR #589, opened 2021-04-12
Refactor Sensor table

CarsonHerness:
This makes the sensor table more readable and more similar to the admin page by separating out active and inactive sensors.

This PR also:

- Move helper components to their own files, and adds folders as needed
- Uses `number` instead of `string` for PurpleAir IDs
- Combines the latitude and longitude columns of the sensor tables to make them more narrow
- Fixes issue with sensor management buttons being too wide on mobile (now they're flexible)


public/locales/es/administration.json
```diff
@@ -114,27 +114,24 @@
         "confirmChange": "Confirma el cambio de status de administrador"
     },
     "sensors": {
-        "heading": "Los sensors actuales",
+        "activeHeading": "Los sensors activos",
```
akampen, line 117:
```suggestion
        "activeHeading": "Los sensores activos",
```

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
```diff
@@ -61,8 +63,8 @@ const DeleteSensorModal: ({sensors}: DeleteSensorModalProps) => JSX.Element = ({
   // --------------- End state maintenance variables ------------------------
 
   const readyToSubmit =
-    purpleAirId !== '' &&
```
akampen, line 64: should this be replaced with a check `purpleAirId !== Number.NaN`, or was this an extraneous check that can be safely removed?

src/components/Admin/ManageSensors/Util/Util.ts
```diff
@@ -0,0 +1,35 @@
+import firebase from '../../../../firebase';
+
+/**
+ *
```
akampen, line 4: nit: delete empty line

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
```diff
@@ -239,7 +241,7 @@ const DeleteSensorModal: ({sensors}: DeleteSensorModalProps) => JSX.Element = ({
               <Box>
                 <LabelValue
                   label={t('sensors.purpleAirId')}
-                  value={purpleAirId}
+                  value={purpleAirId.toString()}
```
akampen, line 244: Should this be replaced with `numberToString()` from Util/Util.ts in case `Number.NaN` every sneaks in?

akampen (approved):

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
CarsonHerness: I basically just checked that the confirmPurpleAirId (which is a string from the user input) is not the empty string and then I compare the purpleAirId to the number converted confirmPurpleAirId

src/components/Admin/ManageSensors/DeleteSensorModal.tsx
CarsonHerness: Yes good catch

PR #590, opened 2021-04-13
Changed fonts

akampen:
This PR:
- changed fonts
- made the graph numbers sans serif (as per Lotenna's request)
- bolded/italicized a few things
- made graph labels black

If you don't like any of the fonts/styles/anything just lmk. I don't feel attached to any, they were curated by Maddy but she is welcome to feedback.


CarsonHerness (approved):
Love it, should we also change the NavBar fonts for consistency? I think we can also set default Chakra font settings that will automatically style headings, etc. in the `AppProviders` file with the `ChakraProvider`

PR #591, opened 2021-04-13
Added jump links on mobile for About and Health Info

akampen:


src/pages/Health.tsx
```diff
@@ -8,17 +9,76 @@ import {
...
+          <Text textAlign="center" fontStyle="italic">
+            Jump to:
+          </Text>
+          <Grid
+            width="100%"
+            templateColumns="repeat(2,1fr)"
+            gap={1}
+            textAlign="center"
+          >
+            <Link gridRow={1} href="/health/#pollution" color="#32bfd1">
```
CarsonHerness, line 67: I think you need to add `id`s to the headings in this file yet

src/pages/Health.tsx
```diff
@@ -8,17 +9,76 @@ import {
...
 const Health: React.FC = () => {
   const {t} = useTranslation('health');
+  const [isMobile, setIsMobile] = useState(
+    window.matchMedia('(max-width: 47.9em)')?.matches ?? false
+  );
+  // -------- Detect screen size for conditional formatting --------- //
+  /**
+   * Adjust UI depending on screenwidth. This function is called from event
+   * listeners with a max-width match media query.
+   * @param this - a media query that either matches or doesn't
```
CarsonHerness, line 26: add note for when "`matches`"

src/pages/Health.tsx
```diff
@@ -8,17 +9,76 @@ import {
...
+  }, []);
+  // -----------------  End detect screen size ----------------- //
 
   return (
     <Flex width="full" align="center" direction="column" padding={8}>
       <Heading as="h1">{t('pageHeading')}</Heading>
+      {isMobile && (
+        <Box>
+          <Text textAlign="center" fontStyle="italic">
+            Jump to:
```
CarsonHerness, line 59: Will need to translate

CarsonHerness (commented):
Looks mostly good but some `id`s are missing from the Health page

src/pages/Health.tsx
akampen: Yep totally forgot, ty

CarsonHerness (commented):
Okay weird issue that I'm seeing: when anything is added after the `health` in the URL like `/` or `/#protection`, then the images don't load--I have no idea what causes this tho

CarsonHerness (approved):
Looks good!

PR #600, opened 2021-04-13
Use SelectedSensor state instead of lots of other states

akampen:
This PR changes the way we pass information from the map to the home page and its sub-components. Now we pass around a `SelectedSensor` which contains info like purpleAirId and sensorDocId instead of keeping track of everything individually. 

Doing this makes the code a bit nicer and also gets rid of the extra database read in `InvalidSensor.tsx`.

It also adds the sensor name to the Gauge UI component. 

Lastly, can you make sure to run it locally and play around on the home page to ensure no unexpected behavior? I've tested it but would love someone to double check.


src/util.ts
```diff
@@ -12,4 +14,18 @@ const aqiCutoffs = {
...
+/**
+ * Interface for a Selected Sensor. This is used in the Home page to keep track
+ * of the values for the sensor that is currently selected on the map.
+ */
+interface SelectedSensor {
+  purpleAirId: number;
+  sensorDocId: string;
+  name: string;
+  aqi: string;
+  isAqiValid: boolean;
```
CarsonHerness, line 26: nit: use `isValid` rather than `isAqiValid` just for consistency throughout the codebase. We could add tsdoc with a list of the values that further clarifies what `isValid` means

src/util.ts
```diff
@@ -12,4 +14,18 @@ const aqiCutoffs = {
...
+  lastValidAqi: firebase.firestore.Timestamp | null;
```
CarsonHerness, line 27: nit: similar, for consistency use `lastValidAqiTime`

CarsonHerness (commented):
I'm getting these errors in the console when I resize the window while looking at the home page. Otherwise, this looks good to me!

image

akampen:
> I'm getting these errors in the console when I resize the window while looking at the home page. Otherwise, this looks good to me!
> 
> image

Opening issue #604 for this. This is unrelated to this PR and is also present on the live website.

src/components/AqiGauge/InvalidSensor.tsx
```diff
@@ -29,43 +29,30 @@ const InvalidSensor: ({purpleAirId}: InvalidSensorProps) => JSX.Element = ({
   // this state variable stores which form of "at" we should use
   const [atString, setAtString] = useState(t('invalid.atPlural'));
```
CarsonHerness, line 30: okay I'm not sure if this is an issue, but strangely the type of `atString` is `() => never`

src/util.ts
```diff
@@ -12,4 +14,24 @@ const aqiCutoffs = {
...
+ * - name: The name of this sensor on PurpleAir
+ * - aqi: The current AQI value of this sensor
+ * - isValid: boolean, is the sensor currently displaying an AQI
+ * - lastValidAqiTime: the last time this sensor reported a valid AQI
+ */
+interface SelectedSensor {
+  purpleAirId: number;
+  sensorDocId: string;
+  name: string;
+  aqi: string;
```
CarsonHerness, line 31: Why is AQI a string?

CarsonHerness (approved):

src/util.ts
akampen: it's always been a string, so I just maintained it because that's how it's being used in other parts of the code. I can change it to a number and update the rest of the codebase to reflect that if you think that'd be better. Thoughts?

src/components/AqiGauge/InvalidSensor.tsx
akampen: something to do with translate, I just added `'' + t('invalid.atPlural')` and now it says it's a string. The functionality works so maybe it's a VScode thing?

src/util.ts
CarsonHerness: Where do you first set the `aqi`? If it's take from Firestore and then immediately converted to a 2-digit string, then I think leaving it as a string makes sense

PR #606, opened 2021-04-14
Store sensor errors for a sensor in each sensor doc

CarsonHerness:
This PR takes care of the backend for the reasons why a sensor might be down. The next PR will implement the front end and show admin users why a sensor might be down.

There are two types of errors: errors with a reading and errors that come up in the AQI calculation process that make a sensor invalid.

I also added a channel downgraded check so that if a channel of a sensor has been downgraded, the mean percent difference for that sensor is automatically set at the maximum, 2.

Finally, I verified that duplicate readings are not be added for sensors.


functions/src/aqi-calculation/cleaned-reading.ts
```diff
@@ -127,24 +136,43 @@ function getHourlyAverages(
...
+          aqiCalculationErrors[InvalidAqiErrors.NotEnoughNewReadings] = true;
+        }
       }
     }
   } else {
     // If no buffer exists, there are no valid readings for any hour
     averages.fill(null);
+
+    // If there is no buffer, then there's not enough new readings
+    aqiCalculationErrors[InvalidAqiErrors.NotEnoughNewReadings] = true;
```
akampen, line 173: Is this the right error for there not being a buffer? It seems like there'd only not be a buffer if the sensor was brand new (or just being re-activated). "Not enough new readings" is totally fine w/ me, it seems to get across the idea that it needs more readings to report AQI, but I also would consider adding the reason "needs more time to collect data" or something that gets across that the error exists b/c it's new.

functions/src/aqi-calculation/invalid-aqi-errors.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Errors that can occur during the AQI calculation process. Any of these errors
+ * will result in a sensor being invalid.
+ *
+ * - `InfiniteAqi`- If the calculated AQI for a sensor is infinite.
+ * - `NotEnoughNewReadings` - If not enough new readings were received for a
+ *   sensor in the last 3 hours. Two of the three most recent hours must have at
+ *   least 23 new (and valid) readings. This error only reflects if not enough
+ *   new readings are being received. This error can also occur soon after a
+ *   sensor is activated if not enough new readings are stored in the PM2.5
+ *   buffer yet.
```
akampen, line 11: This description seems adequate enough to ignore my previous comment about creating a new error for new sensors/

functions/src/aqi-calculation/invalid-aqi-errors.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Returns a default error array for AQI calculation errors
+ * @returns an array for each `InvalidAqiErrors` with each error set to `false`.
+ */
+function getDefaultInvalidAqiErrors(): boolean[] {
+  // TypeScript does not provide a way to get the number of elements in an
+  // enumeration, so this gets the number of elements using the enumeration
+  // reverse mapping.
+  // eslint-disable-next-line no-magic-numbers
+  const invalidAqiErrorCount = Object.keys(InvalidAqiErrors).length / 2;
```
akampen, line 32: why divide by 2?

functions/src/get-reading/purple-air-response.ts
```diff
@@ -113,36 +165,69 @@ function getReading(
...
+     * has a channel downgraded. This is not the behavior we want because we
+     * need both channels' readings to be able to determine if we want to use
+     * that reading in our AQI calculation since we want to be able to know how
+     * confident we can be that a sensor reading is true and not an outlier.
+     * Thus, if either channel of a sensor is downgraded, we set the
+     * meanPercentDifference to the maximum value of 2.
+     */
+    if (
+      sensorErrors[SensorReadingErrors.ChannelADowngraded] ||
+      sensorErrors[SensorReadingErrors.ChannelADowngraded]
```
akampen, line 190: copy paste error, want `ChannelBDowngraded`

functions/src/get-reading/sensor-errors.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Returns a default error array for sensor errors
+ * @returns an array for each `SensorReadingError` with the error set to `false`.
+ */
+function getDefaultSensorReadingErrors(): boolean[] {
+  // TypeScript does not provide a way to get the number of elements in an
+  // enumeration, so this gets the number of elements using the enumeration
+  // reverse mapping.
+  // eslint-disable-next-line no-magic-numbers
+  const sensorReadingErrorCount = Object.keys(SensorReadingErrors).length / 2;
```
akampen, line 40: again, I trust that you're right, but not sure why we divide by 2

functions/src/get-reading/purple-air-response.ts
```diff
@@ -96,6 +140,14 @@ function getReading(
...
         }
         break;
+      case 'channel_flags':
+        if (typeof value === 'number') {
+          const channelDownErrors = getChannelFlag(value, channelFlagNames);
+
+          // Add the channel down errors to the overall sensor errors
+          channelDownErrors.forEach(error => (sensorErrors[error] = true));
+        }
+        break;
```
akampen, line 150: I'm a little lost on what `channel_flags` is. Is it a number? A string array? I don't fully understand what's going on w/ this variable.

akampen (approved):

functions/src/aqi-calculation/cleaned-reading.ts
CarsonHerness: I could make a separate error for this, but the second time we try to calculate the AQI, it will be indistinguishable from "not enough new readings" since the buffer defaults to being filled with nulls. I can make the explanation for this error more clear though and add that "if you just activated or added this sensor, it may also need time to collect enough readings to calculate an AQI"

functions/src/get-reading/purple-air-response.ts
CarsonHerness: oops thank you

functions/src/get-reading/sensor-errors.ts
CarsonHerness: lol it was from Stack Overflow and I didn't question it because it worked on repl.it, I'll investigate and add a comment

functions/src/get-reading/purple-air-response.ts
CarsonHerness: channel_flags is a number that's an index in the channelFlagNames array returned by the PurpleAir API. I'll add a comment explaining this to the docstring of `getChannelFlag`

CarsonHerness:
I think I'm going to change the implementation slightly to be more clear from an array of boleans (where each index of the array corresponds to if that error is true) to an array of integers that if the integer is present, then that error is present. I think this might make it more clear and easier to propagate errors.

PR #610, opened 2021-04-14
Load images from src rather than public

CarsonHerness:
The Spanish/English images for the PM2.5 particles were being loaded by a translation of a filename. Instead, I now use the i18n.language to decide which image to use, and the images are in the media folder rather than the public folder. On my computer, this is faster, and the images correctly load when you navigate to "/health/" (rather than "/health"), which is correct behavior that does not happen on the live website.

I also created a `Health` folder in `components` because I considered making an image helper component for the `Health` page that could be lazy loaded (so that both images aren't loaded into the component). This may still be worth doing, so lmk what you think


akampen (approved):
As far as lazy loading the images w/ a component, if it's easy and faster, sounds great to me. Also seems unessential, so dw about it if you'd rather be done at this point.

PR #625, opened 2021-04-15
Fix deletion function

CarsonHerness:
Changes in this PR:

- Adds `lastUpdated` field to the todo doc in the deletion collection
- Updates how we delete a sensor doc from the deletion map to fix some value not actually being deleted issues
- Adds more awaits when necessary
- Updates batch size to the max batch size, 500


src/components/Admin/ManageSensors/DeleteOldDataModal.tsx
```diff
@@ -131,7 +131,10 @@ function DeleteOldDataModal(): JSX.Element {
         mapping[sensorDoc.id] = deleteBeforeDate;
       }
 
-      deletionDocRef.update({deletionMap: mapping});
+      deletionDocRef.update({
+        deletionMap: mapping,
+        lastUpdated: firebase.firestore.FieldValue.serverTimestamp(),
```
akampen, line 136: lastUpdated refers both to the last time new docs were marked for deletion and the last time docs were deleted. What's the point of the field?

akampen (approved):

src/components/Admin/ManageSensors/DeleteOldDataModal.tsx
CarsonHerness: Similar to why I added it to each of the sensor docs, it's to make sure that some field is updated whenever `.update()` is called, to avoid errors, and then I added it to these fields for consistency. I could also change this to be `lastDeleted` and `mostRecentUpdate` or something like that

PR #626, opened 2021-04-15
Move files and update maximum percent diff

CarsonHerness:
Changes in this PR

- Moves purple-air-response and associated files to its own folder, since it is not doing `aqi-calculation`
- Updates the value returned for a sensor with zero confidence from `NaN` to 2, which is the maximum possible mean percent difference. This change is being made because the `NaN` was showing up in the downloaded sensor reading files as the empty string


functions/src/get-reading/util.ts
```diff
@@ -1,19 +1,12 @@
...
- */
-const readingsSubcollection: (docId: string) => string = (docId: string) =>
-  `/sensors/${docId}/readings`;
-
 /**
  * Converts PurpleAir's confidence value into percent difference
  * @param confidence - confidence value from PurpleAir, between 0 and 100
- * @returns meanPercentDifference value, or NaN if value lost in calculation.
- * Any input that results in NaN means that the meanPercentDifference is high
+ * @returns meanPercentDifference value, or 2 (the maximum possible percent difference ) if the value is lost in calculation.
```
akampen, line 6: What makes 2 the maximum possible percent difference?

akampen (approved):

functions/src/get-reading/util.ts
CarsonHerness: Percent difference is the absolutely value of the difference over the average, like so for values `a` and `b`: `(2*|a-b|)/(a+b)`. When `a` and `b` are the same, the percent difference is `0` since `a-b` is 0. Then consider when `a` and `b` are SUPER different, let's say `a` is tiny and `b` is huge. In the limit as `b` goes to infinity and `a` is epsilon, `|a-b|/(a+b)` goes to 1, so the whole fraction goes to 2.

A fun [Desmos ](https://www.desmos.com/calculator/gxbiwg23ef)to see this as well

PR #632, opened 2021-04-16
Return to Admin Home on top of pages

akampen:
Added "Return to Admin Home" button to the top of Manage Sensors and Manage Users page. Capitalized "out" in "Sign Out".


src/components/Admin/ManageSensors/ManageSensors.tsx
```diff
@@ -80,6 +80,9 @@ const ManageSensors: () => JSX.Element = () => {
           textAlign="center"
         >
           <Heading marginY={2}>{t('manageSensors')}</Heading>
+          <Button as="a" href="/admin" margin={1}>
```
CarsonHerness, line 83: nit: add some padding below the button to separate it from the grid of buttons

CarsonHerness (approved):

PR #643, opened 2021-04-17
Move Activate/Deactivate to button instead of column

CarsonHerness:
This PR is in preparation for adding two more columns with the sensor down reasons.

Changes in this PR:

- Activation/Deactivation now happens through a modal with a button at the top
- All sensor translations were moved to their own file
- Scrolling on the table is fixed! Now all horizontal scrolling in the app is disabled

I'm fairly confident that I updated every `t('sensors....')` so that the correct translations show up, but please click around to make sure I didn't miss anything


src/components/Admin/ManageSensors/ManageSensors.tsx
```diff
@@ -79,32 +79,28 @@ const ManageSensors: () => JSX.Element = () => {
...
+              'repeat(3,1fr)',
+              'repeat(6,1fr)',
+            ]}
             gap={2}
           >
+            <ToggleActiveModal sensors={inactiveSensors} isActive={false} />
             <AddSensorModal />
             <DownloadCSVModal sensors={sensors} />
-            <DeleteOldDataModal />
+            <ToggleActiveModal sensors={activeSensors} isActive />
```
akampen, line 96: Didn't know you could do `isActive` instead of `isActive={true}`. Maybe use the long form for readability?

akampen (approved):

PR #648, opened 2021-04-18
Implement sensor down front end

CarsonHerness:
This PR adds two columns to the sensor table with badges for different error types. When these errors are clicked, an explanation pops up.

Note that the enums are copied directly from the backend.

This PR also:

- Fixes a bug in the backend that resulted in duplicate errors for InvalidAqiError types
- Renames `InvalidAqiErrors` to `InvalidAqiError`


src/components/Admin/ManageSensors/SensorTable/ErrorTags.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * @returns tag for an error to be displayed in the sensor table. When clicked, the tag will show a popover with an explanation of the error.
+ */
+const ErrorTag: ({name, explanation}: ErrorTagProps) => JSX.Element = ({
+  name,
+  explanation,
+}: ErrorTagProps) => {
+  return (
+    <Popover>
+      <PopoverTrigger>
+        <Button colorScheme="red" size="xs">
```
akampen, line 42: nit: consider not red? I know red is  the "error color", so this is totally opinion (and maybe a bad one). I just felt like it might be nice if the error buttons were differentiated from the delete-relates buttons we already have on the page.

src/components/Admin/ManageSensors/SensorTable/ErrorTags.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Portal>
+        <PopoverContent>
+          <PopoverArrow />
+          <PopoverHeader>
+            <Flex alignItems="center" justifyContent="center" marginTop="1em">
+              <WarningTwoIcon color="red.500" />
+              <Heading size="md" color="red.500">
+                {name}
+              </Heading>
+              <WarningTwoIcon color="red.500" />
```
akampen, line 55: Nit: consider adding a space in between the warning icon and the heading.

akampen: There's one tag in the sensor reading that goes onto a second line and it doesn't look good imo with the warning icons.

public/locales/en/sensors.json
```diff
@@ -102,5 +102,49 @@
...
+            "heading": "Channel B Downgraded",
+            "explanation": "Channel B has been downgraded by PurpleAir, likely indicating sensor malfunction."
+        }
+    },
+    "aqiErrors": {
+        "heading": "Recent AQI Calculation Errors",
+        "explanation": "These errors are possible reasons that a sensor does not currently have a valid AQI.",
+        "tooHigh": {
+            "heading": "AQI Too High",
+            "explanation": "The calculated AQI was higher than the possible AQI range, which likely means that the sensor is malfunctioning."
```
akampen, line 139: nit: perhaps change to "the calculated AQI was over 1000, which is not possible" or something like that ("the calculated AQI was infinite.."). I'm just wondering if they get this error and try to communicate with PurpleAir they might run into a communication issue w/ the fact that the AQI doesn't really have a max (like 500 is not actually the max even though they say it is). I'm picturing someone telling PA it's higher than possible (without knowing the number) and PA responding that there is no max val.

akampen: I don't feel strongly about this though, so feel free to ignore.

src/ErrorsTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Returns a default error array for sensor errors
+ * @returns an array for each `SensorReadingError` with the error set to `false`.
+ */
+function getDefaultSensorReadingErrors(): boolean[] {
+  // TypeScript does not provide a way to get the number of elements in an
+  // enumeration, so this gets the number of elements using the enumeration
+  // reverse mapping.
+  // eslint-disable-next-line no-magic-numbers
+  const sensorReadingErrorCount = Object.keys(SensorReadingErrors).length / 2;
```
akampen, line 40: document the '/2'

src/ErrorsTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Returns a default error array for AQI calculation errors
+ * @returns an array for each `InvalidAqiErrors` with each error set to `false`.
+ */
+function getDefaultInvalidAqiErrors(): boolean[] {
+  // TypeScript does not provide a way to get the number of elements in an
+  // enumeration, so this gets the number of elements using the enumeration
+  // reverse mapping.
+  // eslint-disable-next-line no-magic-numbers
+  const invalidAqiErrorCount = Object.keys(InvalidAqiErrors).length / 2;
```
akampen, line 75: document the '/2'

src/components/Admin/ManageSensors/SensorTable/SensorTable.tsx
```diff
@@ -25,6 +37,113 @@ const SensorTable: ({title, sensors}: SensorTableProps) => JSX.Element = ({
...
+  /**
+   * Get the error tags for all sensor errors of a sensor
+   * @param sensor - the sensor to check the sensor errors of
+   * @returns an array of the error tags that are true for that sensor
+   */
+  function getSensorErrorTags(sensor: Sensor): JSX.Element[] {
+    const errorTags: JSX.Element[] = [];
+    sensor.sensorReadingErrors.forEach((value, index) => {
+      if (value) {
+        const [name, explanation] = getSensorReadingErrorInfo(index);
```
akampen, line 94: This seems to be passing a number `index` into `getSensorReadingErrorInfo`, but `getSensorReadingErrorInfo` expects an input parameter `error`. I see how this works, but it's not clear upon a first read. Consider changing this to `error_index` or `error_value` and then edit the documentation for `getSensorReadingErrorInfo` to mention that the input error is a number.

src/components/Admin/ManageSensors/SensorTable/SensorTable.tsx
```diff
@@ -25,6 +37,113 @@ const SensorTable: ({title, sensors}: SensorTableProps) => JSX.Element = ({
...
+        errorTags.push(<ErrorTag name={name} explanation={explanation} />);
+      }
+    });
+    return errorTags;
+  }
+
+  /**
+   * Get the human readable error name and explanation
+   * @param error - the error that occurred
```
akampen, line 103: same documentation comment as above

src/components/Admin/ManageSensors/SensorTable/SensorTable.tsx
```diff
@@ -25,6 +37,113 @@ const SensorTable: ({title, sensors}: SensorTableProps) => JSX.Element = ({
...
+  /**
+   * Get the error tags for all invalid AQI of a sensor
+   * @param sensor - the sensor to check the invalid AQI errors of
+   * @returns an array of the error tags that are true for that sensor
+   */
+  function getInvalidAqiErrorTags(sensor: Sensor): JSX.Element[] {
+    const errorTags: JSX.Element[] = [];
+    sensor.aqiCalculationErrors.forEach((value, index) => {
+      if (value) {
+        const [name, explanation] = getInvalidAqiErrorInfo(index);
```
akampen, line 140: same documentation comment as above

akampen (approved):

src/components/Admin/ManageSensors/SensorTable/ErrorTags.tsx
CarsonHerness: What color do you think would work well?

src/components/Admin/ManageSensors/SensorTable/ErrorTags.tsx
CarsonHerness: For some reason it seems to ignore the spaces but I'll try adding a spacer object or something

public/locales/en/sensors.json
CarsonHerness: Maybe I'll add that the sensor is reporting impossibly high PM2.5 readings

src/components/Admin/ManageSensors/SensorTable/SensorTable.tsx
CarsonHerness: I think updating the error array implementation to a list of integers that correspond to errors (instead of an array of booleans) will make this more clear

src/components/Admin/ManageSensors/SensorTable/ErrorTags.tsx
CarsonHerness: I used an HStack and this worked well

PR #650, opened 2021-04-19
Allow admin to delete non-admin account

CarsonHerness:
This marks a user's doc as `isDeleted` so that if the user tries to sign in before the account is actually deleted, then they know that their account will be deleted. This also adds the user's ID to the deletion fields to delete the Firebase Authentication account and delete the user doc.


src/components/Admin/ManageUsers/DeleteUserPopover.tsx
```diff
@@ -0,0 +1,128 @@
...
+   * @remarks We don't delete the user's document immediately since if the user signs into their account before the Cloud Function deletes their account, their user document will be automatically recreated.
+   */
+  function markUserForDeletion(): Promise<void> {
+    if (isAdmin) {
+      return firestore
+        .collection('deletion')
+        .doc('users')
+        .update({
+          userDocs: firebase.firestore.FieldValue.arrayUnion(user.userId),
+          firebaseUsers: firebase.firestore.FieldValue.arrayUnion(user.userId),
```
akampen, line 69: update `lastUpdated`

akampen (approved):

PR #656, opened 2021-04-19
Add AccountDeleted Page

CarsonHerness:
When a user's account has been marked as `isDeleted` (which happens when an admin deletes a user's account), there's a period of 24 hours where the user could still be able to log in even though the account is deleted. This PR adds a page that's displayed to a user in this case.

You can test this PR locally by manually changing one of your non-admin accounts to `isDeleted=true` in Firestore. The next PRs will include functionality to set this as true.

This PR also:

- Moves the ToggleUserPopover to its own component
- Updates SignOut because I was getting some compilation errors
- Updates the user page headings to be "h2" elements

### DIFF
--- a/public/locales/en/administration.json
+++ b/public/locales/en/administration.json
@@ -51,7 +51,6 @@
         "email": "Sign up with email"
     },
     "signOut": "Sign Out",
-    "signOutError": "Error occurred with sign out. Please try again",
     "pageHeader": {
         "signIn": "Sign In",
         "signUp": "Sign Up"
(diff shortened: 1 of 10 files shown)


public/locales/en/administration.json
```diff
@@ -112,5 +111,9 @@
         },
         "changeAdminError": "Unable to change admin status for ",
         "confirmChange": "Confirm Admin Status Change"
+    }, 
+    "deletedAccount": {
+        "deleteNow": "Finish Account Deletion Now",
+        "explanation": "Your account has been deleted by an administrator of the website, and the account will be deleted within the next 24 hours. If you believe this is a mistake, reach out to an administrator of the website and recreate your account after this account has been deleted. You may choose to complete account deletion at this time."
```
akampen, line 117: What does this mean?: "You may choose to complete account deletion at this time."

akampen: nvm, change text to make it more clear that you do this by "clicking the button"

src/components/Admin/AccountDeleted.tsx
```diff
@@ -0,0 +1,40 @@
...
+        padding={8}
+        marginX={8}
+        width="full"
+        maxWidth="500px"
+        borderWidth={1}
+        borderRadius={8}
+        boxShadow="lg"
+        textAlign="center"
+      >
+        <Text>{t('deletedAccount.explanation')}</Text>
```
akampen, line 30: Add heading of some kind (ex: "Account Deleted" or "Notice")

PR #657, opened 2021-04-19
Link Colors

akampen:
I added a `LinkColor` to the colors and used that for all the links on the site. Also made the links slightly darker, but if you have a different color preference feel free to text me the hex or change it yourself, I'm not attached to it.


CarsonHerness (approved):
Looks good, might I suggest a color: `#0081de`. Also, `ForgotPassword`, `SignIn` and `SignUp` all have links that still use `teall.500` but they're for the sign in/sign up pages, so maybe they don't need to use `LinkColor`? wdyt

PR #662, opened 2021-04-20
Update Firestore rules to check fields updated

CarsonHerness:
If a user deletes their own account after an admin has marked their account for deletion, we have to remove the uid from the firebaseUsers list in the deletion collection. Otherwise, the deletion cloud function tries to delete a Firebase user that does not exist.

These Firestore rules allow these sorts of changes. For non-admin users, these rules check that only one field is modified and that the modification is either adding or removing their own user ID.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -68,26 +68,33 @@ service cloud.firestore {
 
     // Used for deletion of user data
     match /deletion/users {
-      // Verify that a (1) user is signed in, (2) the only modified field is
+      let newData = request.resource.data;
+      let oldData = resource.data;
+
+      // Verify that (1) user is signed in, (2) the only modified field is
       // userDocs, (3) the userDocs array is increased in size by only 1,
       // and (4) the added item from the userDocs array is the user's uid.
       function addUidToUserDocs() {
-        let onlyUserDocsUpdated = request.resource.data.keys().hasOnly('userDocs');
-        let oneItemRemoved = request.resource.data.users.size() == resource.data.users.size() + 1;
-        let addedUid = request.resource.data.users.removeAll(resource.data.users)[0] == request.auth.uid;
+        // The diff() function combined with affectedKeys() returns a list of
+        // keys that were changed in an update operation
+        let onlyUserDocsUpdated = newData.diff(oldData).affectedKeys().hasOnly('userDocs');
+        let oneItemAdded = newData.userDocs.size() == oldData.userDocs.size() + 1;
+        let addedUid = newData.userDocs.removeAll(oldData.userDocs)[0] == request.auth.uid;
         return isSignedIn()
                 && onlyUserDocsUpdated
-                && oneItemRemoved
+                && oneItemAdded
                 && addedUid;
       }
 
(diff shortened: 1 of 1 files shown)


firestore.rules
```diff
@@ -63,14 +68,35 @@ service cloud.firestore {
 
     // Used for deletion of user data
     match /deletion/users {
-      function markingCurrentUserDoc() {
-        // Verify that the current user is adding their user ID to the userDocs field
-        return isSignedIn() && request.auth.uid in request.resource.data.userDocs;
+      // Verify that a (1) user is signed in, (2) the only modified field is
```
akampen, line 71:
```suggestion
      // Verify that (1) the user is signed in, (2) the only modified field is
```

firestore.rules
```diff
@@ -63,14 +68,35 @@ service cloud.firestore {
...
+      // userDocs, (3) the userDocs array is increased in size by only 1,
+      // and (4) the added item from the userDocs array is the user's uid.
+      function addUidToUserDocs() {
+        let onlyUserDocsUpdated = request.resource.data.keys().hasOnly('userDocs');
+        let oneItemRemoved = request.resource.data.users.size() == resource.data.users.size() + 1;
+        let addedUid = request.resource.data.users.removeAll(resource.data.users)[0] == request.auth.uid;
```
akampen, line 77: this is a little hard to understand, consider something like `let removedUid = newData.removeAll(oldData)[0] = request.auth.uid;` where `oldData = resource.data.users` and `newData = request.resource.data.users`

firestore.rules
```diff
@@ -63,14 +68,35 @@ service cloud.firestore {
...
+        return isSignedIn()
+                && onlyUserDocsUpdated
+                && oneItemRemoved
+                && addedUid;
+      }
+
+      // Verify that a (1) user is signed in, (2) the only modified field is
```
akampen, line 84:
```suggestion
      // Verify that (1) the user is signed in, (2) the only modified field is
```

firestore.rules
```diff
@@ -63,14 +68,35 @@ service cloud.firestore {
...
+      // firebaseUsers, (3) the firebaseUsers array is reduced in size by only 1,
+      // and (4) the removed item from the firebaseUsers array is the user's uid.
+      function removeUidFromFirebaseUsers() {
+        let onlyFirebaseUsersUpdated = request.resource.data.keys().hasOnly('firebaseUsers');
+        let oneItemRemoved = request.resource.data.users.size() == resource.data.users.size() - 1;
+        let removedUid = resource.data.users.removeAll(request.resource.data.users)[0] == request.auth.uid;
```
akampen, line 90: this is a little hard to understand, consider something like `let removedUid = oldData.removeAll(newData)[0] = request.auth.uid;` where ` oldData = resource.data.users` and `newData = request.resource.data.users`

akampen (commented):
If possible you could also use something like `oldData` and `newData` for the other `request.resource....`, but not necessary if you think it's good w/o

akampen (approved):

PR #665, opened 2021-04-20
Add ability to add a password

CarsonHerness:
This is for users who signed up using Google. Adding a password allows a user (in the future) to change their email, since we won't allow non-password based users to change their email.

I also am dumb and forgot to update the reauth component to use context, so this PR does that and adds a `form` tag around the confirm password field.


src/components/Admin/ManageAccount/AddPassword.tsx
```diff
@@ -0,0 +1,200 @@
...
+        {t('addPassword.heading')}
+      </Button>
+      <Modal isOpen={isOpen} onClose={handleClose}>
+        <ModalOverlay />
+        <ModalContent>
+          <ModalHeader>{t('addPassword.heading')}</ModalHeader>
+          <ModalCloseButton />
+          <ModalBody>
+            <Text>{t('addPassword.explanation')}</Text>
+            <Reauthentication
```
akampen, line 151: Add some space between the Reauth component and the add password explanation.

src/components/Admin/ManageAccount/AddPassword.tsx
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Resets all possible errors to no error.
+   */
+  function resetErrors() {
+    setNewPasswordError('');
+    setConfirmNewPasswordError('');
+    setError('');
+  }
+
+  function addPassword(): Promise<void> {
```
akampen, line 86: add docstring

akampen (approved):

PR #666, opened 2021-04-20
Convert google and password user to context

CarsonHerness:
This allows `googleUser` and `passwordUser` to be accessed more easily rather than passing the values between components


akampen (approved):
Carson stop fixing things that don't fix things that are broken.

CarsonHerness:
> Carson stop fixing things that don't fix things that are broken.

Okay it was slowing me down when trying to add the other account update stuff, I swear it's worth it

PR #674, opened 2021-04-20
Fix Firestore Rules

CarsonHerness:
This time, I checked that the rules compiled in Firestore and I accidentally deployed them there too.

This [blog post](https://dev.to/chinchang/restrict-specific-fields-updation-in-firebase-firestore-ohp) was really helpful for figuring out how to restrict updates to a field

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -1,6 +1,11 @@
 rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
+    // Note that request.resource.data contains the incoming data update while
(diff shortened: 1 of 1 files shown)


akampen (approved):
Everything seems to be indented one block. Probably desired, just wanted to point it out.

CarsonHerness:
> Everything seems to be indented one block. Probably desired, just wanted to point it out.

Yes, this PR also nests the `users` doc check and the `readings` doc check under the `match` for the `deletion` collection, instead of doing them separately

PR #675, opened 2021-04-20
implemented help icons on home page

akampen:
User testing revealed that AQI was not defined in enough places. A help icon with a short, functional definition of AQI can be found in three places:
- Dial component before a sensor is clicked
- Dial component after a sensor is clicked (next to the AQI itself)
- Map component after a sensor is clicked (in the graph title)


src/components/Util/MoreInfoLabel.tsx
```diff
@@ -0,0 +1,94 @@
...
+  fontSize,
+  fontFamily,
+}: MoreInfoLabelProps) => JSX.Element = ({
+  message,
+  text,
+  fontWeight,
+  fontSize,
+  fontFamily,
+  popoverLabel,
+}: MoreInfoLabelProps) => {
```
CarsonHerness, line 54: Add default values here
```suggestion
  message,
  text,
  fontWeight = 'regular',
  fontSize = 'lg',
  fontFamily = 'Oxygen',
  popoverLabel,
}: MoreInfoLabelProps) => {
```

CarsonHerness (approved):
Looks great! Just a small comment on default values for components with optional parameters

PR #676, opened 2021-04-20
Now cast link

### LINKED ISSUE #640: Add AQI definition to both About and Health pages
I think the About page should also add an explanation that we use the NowCast AQI and link to that

akampen:
I added a brief description and link to the NowCast algorithm to the about page under "Sensor Not Responding?".

I opted not to add any reference to NowCast on the "health information" page because I think that section is targeted towards users with less technical knowledge, and as such adding reference to NowCast detracts from the simple information presented there.

Closes #640


akampen:
I've sent the translations to Hillary on Slack for review. Upon approval, I will not merge until I've heard back from her.

CarsonHerness (approved):
Looks good! I agree with your assessment on the About page.

PR #680, opened 2021-04-21
Scrolls to gauge when new sensor is clicked on mobile

### LINKED ISSUE #549: On mobile, jump to graph when sensor is clicked
Currently it just highlights the sensor and you have to scroll and open a new box--if I were not aware of the graph existing, I would not have known to scroll.

akampen:
Closes #549 

This feature does not work the very first time a sensor is clicked, which is unfortunate, but seems to work well after that point. Please test on your phone as well as mobile and desktop view on your computer to make sure it's acting right!


CarsonHerness:
For me, it defaults to the "#aqiGraph" when I load the page initially, which is a little confusing, but I'm not sure how to change that.

Also, do you think defaulting to having a sensor selected would fix part of the "click twice" thing to go to the graph?

akampen:
> For me, it defaults to the "#aqiGraph" when I load the page initially, which is a little confusing, but I'm not sure how to change that.
> 
> Also, do you think defaulting to having a sensor selected would fix part of the "click twice" thing to go to the graph?

It's possible that defaulting to a selected sensor would fix that, but I think getting rid of the instructions is worse than the page reloading. The instructions feel important since there is so little text on the site. I'll check out the #aqiGraph thing

CarsonHerness:
> 
> It's possible that defaulting to a selected sensor would fix that, but I think getting rid of the instructions is worse than the page reloading. The instructions feel important since there is so little text on the site. I'll check out the #aqiGraph thing

I agree that instructions are important, what do you think about adding instructions into to one of the components or on the map? or above the map?

akampen:
> For me, it defaults to the "#aqiGraph" when I load the page initially, which is a little confusing, but I'm not sure how to change that.
> 
I think this has been fixed, in the latest PR, but it still shows up as soon as you've clicked one sensor.

akampen:
> I agree that instructions are important, what do you think about adding instructions into to one of the components or on the map? or above the map?

I don't **love** the idea of changing the UI that significantly without talking to the liaisons, which makes me feel like we won't have enough time implement that before code freeze.

CarsonHerness:
> 
> I don't **love** the idea of changing the UI that significantly without talking to the liaisons, which makes me feel like we won't have enough time implement that before code freeze.

I was thinking just adding the instruction above the dial and sensor name? But also completely valid to not want to change that at this point

CarsonHerness (approved):

PR #681, opened 2021-04-21
Improve website behavior when offline

akampen:
This PR creates an offline home page and an online home page. When your network status changes, it changes the status of the page. The 10 second delay is there because Here maps throws an error if it can't connect to the network, so automatically switching the component as soon as network connectivity changes doesn't give Here maps enough time to do its thing.

Please test this! It works well on my computer, but does not work at all on mobile (Safari automatically tells me I'm offline, which I tested by disconnecting from my wifi and disabling cellular). We should decide if this fix is an improvement or not.


CarsonHerness (approved):
Looks good! I think this only works if you lose connection while using the website (not if you are offline and then open the website), but I will test that functionality once it gets merged.

PR #682, opened 2021-04-21
Default to a sensor when the page opens

### LINKED ISSUE #634: Default to sensor selected on map
This will help avoid the map reloading the first time someone clicks on a sensor (something users found kinda confusing)

I think we could overlay the instructions to click on a sensor on the map, if possible, because the instructions are still helpful. If not the map, just below it?

akampen:
I kept the default instructions on the home page to display while the map is loading (which for me is too fast to read it, but nice for people with weaker connection/slower browsers).

This closes #634

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -9,5 +9,6 @@
     "moreInformation": "More Information",
     "jumpTo": "Jump to",
     "aqiHelpHeading": "What Is AQI?",
-    "aqiHelpMessage": "AQI stands for Air Quality Index. It is the national standard for indicating how good or bad the air is. Lower numbers indicate safer air, while higher numbers indicate more hazardous air."
+    "aqiHelpMessage": "AQI stands for Air Quality Index. It is the national standard for indicating how good or bad the air is. Lower numbers indicate safer air, while higher numbers indicate more hazardous air.",
+    "noSensorGauge": "Click a sensor on the map to see its AQI value here!"
 }
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -10,5 +10,5 @@
     "jumpTo": "Jump to",
     "aqiHelpHeading": "What Is AQI?",
     "aqiHelpMessage": "AQI stands for Air Quality Index. It is the national standard for indicating how good or bad the air is. Lower numbers indicate safer air, while higher numbers indicate more hazardous air.",
-    "noSensorGauge": "Click a sensor on the map to see its AQI value here!"
+    "instructions": "Click a sensor on the map to see its AQI value here!"
 }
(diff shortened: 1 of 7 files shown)


akampen:
For me this adds some scroll to the home page. Any preferences or suggestions or do you think the added scroll is fine (given that without this PR the scroll is already present depending on window height, which is hard for us to control for)

public/locales/es/common.json
```diff
@@ -9,5 +9,6 @@
     "moreInformation": "Más información",
     "jumpTo": "Salta a",
     "aqiHelpHeading": "¿Qué es AQI?",
-    "aqiHelpMessage": "AQI es el índice de la calidad del aire. Es el estándar nacional para indicar si el aire está bien o mal. Los números más bajos indican aire más saludable, mientras que los números más altos indican aire más peligroso."
+    "aqiHelpMessage": "AQI es el índice de la calidad del aire. Es el estándar nacional para indicar si el aire está bien o mal. Los números más bajos indican aire más saludable, mientras que los números más altos indican aire más peligroso.",
+    "noSensorGauge": "¡Haz clic en un sensor en el mapa para ver el Índice de la calidad de aire (Air Quality Index, AQI) aquí!"
```
CarsonHerness, line 13: In Spanish, the message is pretty long, so I think we can say this since the question mark is right next to it and gives AQI a full definition there
```suggestion
    "noSensorGauge": "¡Haz clic en un sensor en el mapa para ver el AQI aquí!"
```

src/components/Map/Map.tsx
```diff
@@ -188,65 +188,96 @@ class Map extends React.Component<MapProps> {
...
+            const data = firstSensor.getData();
+            const selectedSensor: SelectedSensor = {
+              purpleAirId: data.purpleAirId,
+              sensorDocId: data.sensorDocId,
+              name: data.name,
+              aqi: data.aqi,
+              isValid: data.isValid,
+              lastValidAqiTime: data.lastValidAqiTime,
+            };
+            this.props.updateSelectedSensor(selectedSensor);
```
CarsonHerness, line 213: For some reason when the page first loads, the selected sensor doesn't have the bold black border on it

public/locales/en/common.json
```diff
@@ -9,5 +9,6 @@
     "moreInformation": "More Information",
     "jumpTo": "Jump to",
     "aqiHelpHeading": "What Is AQI?",
-    "aqiHelpMessage": "AQI stands for Air Quality Index. It is the national standard for indicating how good or bad the air is. Lower numbers indicate safer air, while higher numbers indicate more hazardous air."
+    "aqiHelpMessage": "AQI stands for Air Quality Index. It is the national standard for indicating how good or bad the air is. Lower numbers indicate safer air, while higher numbers indicate more hazardous air.",
+    "noSensorGauge": "Click a sensor on the map to see its AQI value here!"
```
CarsonHerness, line 13: Since `noSensorGauge` is no longer the case here, maybe let's call is `instructions` or something like that

CarsonHerness (approved):

PR #683, opened 2021-04-22
Replaced logo and favicon

CarsonHerness:
Let me know if you have artistic criticisms. You may have to run `npm run serve` (instead of `start`) and force refresh to see the changes.


akampen (approved):
Wow that's a lot of files. I'm assuming that you know what you're doing here (cause I don't want to research exactly what icon sizes we need and where we put them). From what I can tell, everything looks good.

CarsonHerness:
> Wow that's a lot of files. I'm assuming that you know what you're doing here (cause I don't want to research exactly what icon sizes we need and where we put them). From what I can tell, everything looks good.

Yeah that was my reaction too. I used https://realfavicongenerator.net/ and it seems that they just handle more icons for more browsers.

PR #685, opened 2021-04-22
Use Reauthentication component

CarsonHerness:
This PR removes the old reauthentication function and uses the new component everywhere.


akampen (commented):
When I press the submit button on the Reauth component in change password, it refreshes the entire page (so it's impossible to change my password).

CarsonHerness:
> When I press the submit button on the Reauth component in change password, it refreshes the entire page (so it's impossible to change my password).

The reauthentication component was accidentally located inside of the change password `form` tag, but I believe the bug is fixed now

PR #688, opened 2021-04-22
Only show error tags for active sensors

CarsonHerness:
The error tags are only relevant for active sensors, and it'd be confusing to show them for inactive sensors since the errors aren't updated


akampen (approved):
I agree! Good change

